Issue #2, opened 2019-07-23
cat_generator.py/inf_generator.py don't support VS2015


ericjin:
The code logic in cat_generator.py/inf_generator.py under edk2toollib/windows/capsule don't support VS2015, and only VS2017 or later. This is a back compatibility issue, the code use these two files(previous version) doesn't work now.

Btw, 3 binaries(vswhere.exe, signtool.exe and inf2cat.exe) are touched at least. The first comes from VS2017, 2nd comes from Windows 10 Software Development Kit (SDK) which will be installed if Desktop development with C++ is selected during VS2017 installation.                                                                                                                                                                                                                  But the 3rd one comes from WDK which should be installed separately. 
For the content on web link https://microsoft.github.io/mu/CodeDevelopment/prerequisites/#visual-studio-2017, the WDK may be missed.

After installation, both WDK and SDK will be installed in C:\Program Files (x86)\Windows Kits\10\bin. 
Signtool.exe is located in C:\Program Files (x86)\Windows Kits\10\bin\10.0.17763.0\x86 and C:\Program Files (x86)\Windows Kits\10\bin\10.0.17763.0\x64
Inf2cat.exe is located in C:\Program Files (x86)\Windows Kits\10\bin\10.0.17763.0\x86   - Question1 - is it correct behavior or I miss something ? no inf2cat.exe under x64.
Question2 -  Should SDK and WDK version be consistent (in the same10.X.X.X ) and/or in the top/highest version?

PR #3, opened 2019-07-24
Add best effort support for locating WinSDK tools

spbrogan:
For platforms that don't have VS2017 or newer installed the vswhere based solution for locating the WinSDK dir doesn't work.  This change adds support for a best effort to look in known default install directories for the WinSDK tools.   The prefered method is still to use VsWhere and the vsvars files but this can help on older platforms.

### DIFF
--- a/edk2toollib/windows/locate_tools.py
+++ b/edk2toollib/windows/locate_tools.py
@@ -1,9 +1,11 @@
 # @file locate_tools.py
 # This module provides python services that locate common development tools using vswhere.exe,
-# vsvars.bat, and other Windows based tools.  This will only work on systems running Windows
-# and with dev tools installed.
+# vsvars.bat, and other Windows based tools.  This works best on systems running Windows
+# and with dev tools installed but will attempt to use known paths for WinSDK if the dev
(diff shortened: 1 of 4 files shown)


spbrogan:
Resolves Issue #2

edk2toollib/windows/locate_tools.py
```diff
@@ -135,11 +137,16 @@ def FindWithVsWhere(products="*"):
...
+    if product is None:
+        product = "*"
+    if arch is None:
+        # TODO: look up host architecture?
+        arch = "amd64"
     interesting = set(keys)
     result = {}
-    ret, vs_path = FindWithVsWhere()
+    ret, vs_path = FindWithVsWhere(product)
     if ret != 0:
```
matthewfcarlson: You need to check if vs_path is not none. It looks like we don't return a non zero ret in all cases

spbrogan: done

ericjin:
This change makes the solution friendly to user who don't install VS2017 or later (just try on one desktop with VS2015), give the hint "We didn't find VS path or otherwise failed to invoke vsWhere".

PR #5, opened 2019-08-01
Add support for limiting vswhere to certain versions of visual studio

spbrogan:


edk2toollib/windows/locate_tools.py
```diff
@@ -37,6 +37,12 @@
 __URL = "https://github.com/microsoft/vswhere/releases/download/{}/vswhere.exe".format(__VERSION)
 __SHA256 = "10abd21aeb5003d87c01f033fd7c170360e362be397f23b0b730324abbd92612"
 
+#
+# Supported Versions that can be queried with vswhere
+# Use lower case for key as all comparisons will be lower case
+#
+supported_vs_versions = {"vs2017": "15.0,16.0", "vs2019": "16.0,17.0"}
```
corthon: Do we want to do any of this as "greater than"? Otherwise we're in the basecore trap of having to rev and release this Pip module every time a new VS is out.

spbrogan: This is only if you want to limit your search for these versions.  I don't think there is an open ended way to do this.

PR #10, opened 2019-08-07
[Bug Fix] Better junit reporting

matthewfcarlson:
This augments a few things in the Junit output based on this: https://github.com/kyrus/python-junit-xml/blob/master/junit_xml/__init__.py This also sanitizes output into the XML

### DIFF
--- a/edk2toollib/log/junit_report_format.py
+++ b/edk2toollib/log/junit_report_format.py
@@ -10,6 +10,7 @@
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 ##
 import time
+from xml.sax.saxutils import escape
 
 
 class JunitReportError(object):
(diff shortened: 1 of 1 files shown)


matthewfcarlson:
@spbrogan I think might help catch future errors in XML by parsing output for < and > which was causing problems before. I think this could go even farther

edk2toollib/log/junit_report_format.py
```diff
@@ -76,15 +77,17 @@ def SetSkipped(self):
...
 
     def LogStdError(self, msg):
-        self.StdErr += msg.strip() + "\n "
+        self.StdErr += escape(msg.strip()) + "\n "
 
     def Output(self, outstream):
         outstream.write('<testcase classname="{0}" name="{1}" time="{2}">'.format(self.ClassName, self.Name, self.Time))
         if self.Status == JunitReportTestCase.SKIPPED:
-            outstream.write('<skipped />')
+            outstream.write('<skipped message="skipped test because of configuration" type="skipped">')
```
spbrogan: I think i would suggest changing this.  A message like this isn't helpful as it is the same for every skipped case.  If you want to support adding a message then i think it needs to be fully plumbed thru so it can be caller controlled.

matthewfcarlson: Removed the message

PR #15, opened 2019-08-16
[Bug Fix] Fix typo in RunPythonScript parameters

matthewfcarlson:
an x was in front of raise_exception_on_nonzero, which caused an exception

### DIFF
--- a/edk2toollib/utility_functions_test.py
+++ b/edk2toollib/utility_functions_test.py
@@ -1,0 +1,24 @@
+# @file utility_functions_test.py
+# unit test for locate_tools module.  Only runs on Windows machines.
+#
+#
+# Copyright (c) Microsoft Corporation
+#
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+import unittest
+import sys
+import os
+import edk2toollib.utility_functions as utilities
+
+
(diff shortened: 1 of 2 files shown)


spbrogan:
Can you write a unit test for this function so that these types of things are caught.

edk2toollib/utility_functions_test.py
```diff
@@ -0,0 +1,24 @@
+# @file utility_functions_test.py
+# unit test for locate_tools module.  Only runs on Windows machines.
```
spbrogan: clean up

edk2toollib/utility_functions_test.py
```diff
@@ -0,0 +1,24 @@
...
+
+import unittest
+import sys
+import os
+import edk2toollib.utility_functions as utilities
+
+
+class UtilityFunctionsTest(unittest.TestCase):
+
+    def test_RunPythonScript(self): # simple run yourself!
```
spbrogan: should we add a test case that uses all named parameters

matthewfcarlson: Added a few more test cases

edk2toollib/utility_functions_test.py
matthewfcarlson: Cleaned up

PR #28, opened 2019-09-17
Fix conditional processing in the base parser

### LINKED ISSUE #25: DSC parser can't handle if TRUE
Exception in parser:
CRITICAL - EXCEPTION: ('Invalid conditional', '!if FALSE')

This also will be for FDF parser

matthewfcarlson:
fixes #25


spbrogan:
can i request that we break the test case into individual test cases so we can see all the different types of conditional support.  In many ways our tests are documentation indicating what we expect our parser to support so making this more clear would be helpful.

matthewfcarlson:
> can i request that we break the test case into individual test cases so we can see all the different types of conditional support. In many ways our tests are documentation indicating what we expect our parser to support so making this more clear would be helpful.

This is a good point- I'll break out the test

edk2toollib/uefi/edk2/parsers/base_parser_test.py
```diff
@@ -0,0 +1,115 @@
...
+        self.assertEqual(len(parser.ConditionalStack), 0)
+
+    def test_process_conditional_ands_ors(self):
+        parser = BaseParser("")
+
+        # make sure we can't do 5 tokens
+        with self.assertRaises(RuntimeError):
+            parser.ProcessConditional("!if 3 == 6 ==")
+
+        # TODO: check for and and if once we've implemented this
```
spbrogan: why is this test called ands_ors?

matthewfcarlson: Because it will test the and and or operators once we support them

edk2toollib/uefi/edk2/parsers/base_parser.py
```diff
@@ -266,27 +285,45 @@ def ProcessConditional(self, text):
         tokens = text.split()
         if(tokens[0].lower() == "!if"):
             # need to add support for OR/AND
-            if(len(tokens) < 4):
+            if (len(tokens) == 2):
+                value = self.ConvertToInt(tokens[1].strip())
+                self.PushConditional(value == 1)  # if the value is true
+            # we can have tokens in 4, 8, 12 etc
+            elif len(tokens) >= 4 and len(tokens) % 4 == 0:
+                con = self.ComputeResult(tokens[1].strip(), tokens[2].strip(), tokens[3].strip())
```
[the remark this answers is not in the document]
matthewfcarlson: Right if it's more than 4 and in multiple of 4 then we can parse it. We need to implement a proper parser with maybe a stack since we need to support () and other things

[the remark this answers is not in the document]
matthewfcarlson: if TEST1 == TEST2 AND test3 == test4 is 8 tokens
if TEST1 == TEST2 AND test3 == test4 OR test5 == test6 is 8 tokens
and so on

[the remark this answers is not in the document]
matthewfcarlson: I didn't add support for +4 tokens but made sure it didn't crash on longer statements

matthewfcarlson:
Things to add:
- all the operators
- Guid parsing
- Section parsing
- Setting variables
- Setting workspace
- broken conditionals for !endif, !else, !ifndef
- replace variables
- parsing hex values
- Find Path

matthewfcarlson:
Let's add in and and or processing as well, why not :)

edk2toollib/uefi/edk2/parsers/base_parser_test.py
```diff
@@ -43,32 +40,80 @@ def test_process_conditional_single_boolean(self):
...
             parser.ProcessConditional("!if 3 ==")
 
-        # check 4 tokens
+    def test_process_conditional_hex_number(self):
+        parser = BaseParser("")
+        # check that a hex number doesn't exqual itself
+        self.assertTrue(parser.ProcessConditional("!IF 0x30 == 30"))
+        self.assertFalse(parser.InActiveCode())
+        self.assertTrue(parser.ProcessConditional("!endif"))
+        self.assertTrue(parser.ProcessConditional("!IF 0x20 == 0x30"))
```
spbrogan: should there be a passing case like 0x20 == 32?

edk2toollib/uefi/edk2/parsers/base_parser_test.py
```diff
@@ -43,32 +40,80 @@ def test_process_conditional_single_boolean(self):
...
-        parser = BaseParser("")
-
         # make sure we can't do three tokens
         with self.assertRaises(RuntimeError):
             parser.ProcessConditional("!if 3 ==")
 
-        # check 4 tokens
+    def test_process_conditional_hex_number(self):
+        parser = BaseParser("")
+        # check that a hex number doesn't exqual itself
```
spbrogan: spelling exqual???

edk2toollib/uefi/edk2/parsers/base_parser_test.py
```diff
@@ -43,32 +40,80 @@ def test_process_conditional_single_boolean(self):
...
+        self.assertTrue(parser.InActiveCode())
         self.assertTrue(parser.ProcessConditional("!IF 30 > 50"))
         self.assertFalse(parser.InActiveCode())
-        # check else
-        self.assertTrue(parser.ProcessConditional("!else"))
+        parser.ProcessConditional("!endif")
         self.assertTrue(parser.InActiveCode())
-        self.assertTrue(parser.ProcessConditional("!EnDiF"))
+        self.assertTrue(parser.ProcessConditional("!IF 30 > 30"))
+        self.assertFalse(parser.InActiveCode())
```
spbrogan: again don't we need a passing case

edk2toollib/uefi/edk2/parsers/hash_file_parser_test.py
```diff
@@ -0,0 +1,45 @@
+# @file guid_parser_test.py
+# Contains unit test routines for the guid parser class.
```
spbrogan: incorrect comment

edk2toollib/uefi/edk2/parsers/base_parser_test.py
matthewfcarlson: Looks like that doesn't work. I think we might have a bug :p

edk2toollib/uefi/edk2/parsers/base_parser_test.py
matthewfcarlson: Good point

edk2toollib/uefi/edk2/parsers/base_parser_test.py
matthewfcarlson: Good spot

matthewfcarlson:
What about $(UNRESOLVED_VARIABLE) == FALSE

matthewfcarlson:
The test cases need better names

matthewfcarlson:
add a test case in your PR for both styles of syntax ifdef using $(MACRO) and ifdef MACRO.

PR #29, opened 2019-10-04
[Bug fix] catch errors on markdown emit

matthewfcarlson:


edk2toollib/log/markdown_handler.py
```diff
@@ -61,6 +61,8 @@ def handle(self, record):
             self.acquire()
             try:
                 self.emit(record)
+            except Exception:  # silently fail
```
spbrogan: would it be better to emit some message to saying we got invalid characters.....agree we should not crash

matthewfcarlson: The downside it we're in this error state because we can't emit. I don't want to log something in case we're having serious problems- that's just infinite and really annoying recursion. I think we should print to console if we want to do an error message.

matthewfcarlson: @spbrogan do we want to print something out here?

PR #32, opened 2019-10-12
Introduce IVRS parser to pytool

kuqin12:
Added ivrs parser and its tests. This tool can be run to parse IVRS acpi table on applicable devices.


matthewfcarlson:
Also a readme for the IVRS parser 😆

edk2toollib/acpi/ivrs_parser.py
```diff
@@ -0,0 +1,643 @@
...
+import sys
+import struct
+import xml.etree.ElementTree as ET
+
+IVRSParserVersion = '1.00'
+
+
+class IVRS_TABLE(object):
+    # Header Lengths
+    IVRSHeaderLength = 48
```
corthon: Generally, these should be calculated from structure strings. That way the reader can easily see what fields contribute to the length and/or modify as needed.

edk2toollib/acpi/ivrs_parser.py
```diff
@@ -0,0 +1,643 @@
...
+    IVHDHeader10HLength = 24
+    IVHDHeaderExLength = 40
+    DeviceTableEntryLengthShort = 4
+    DeviceTableEntryLengthLong = 8
+    DeviceTableEntryLengthVar = 22
+
+    def __init__(self, data):
```
corthon: For each of the structures like this, let's standardize on this interface:
Encode(): Object -> Bytes string
Decode(): Bytes string -> Object fields

Eliminate the __str__ functions and change them to DumpInfo(). DumpInfo() should be patterned like https://github.com/tianocore/edk2-pytool-library/blob/83ea544be8d4a9225fb2c4caac8fc264558df2d0/edk2toollib/uefi/fmp_capsule_header.py#L279

corthon: Also, in the __init__ function, let's make the "data" field optional and just invoke the Decode() function if it's provided.

edk2toollib/acpi/ivrs_parser.py
```diff
@@ -0,0 +1,643 @@
...
+
+    def __str__(self):
+        retval = str(self.ivrs_table)
+
+        for sub in self.ivrs_table.SubStructs:
+            retval += str(sub)
+
+        return retval
+
+    def toBytes(self):
```
corthon: Let's make this the Encode() function, and return just a b'' instead of a "bytearray".

edk2toollib/acpi/ivrs_parser.py
```diff
@@ -0,0 +1,643 @@
...
+                sum += ele
+        else:
+            if (val & 0xFFFFFFFF00000000) != 0:
+                raise Exception("Invalid 32 bit value: %x", val)
+            else:
+                sum = ((val & 0xFF) + ((val >> 8) & 0xFF) + ((val >> 16) & 0xFF) + ((val >> 24) & 0xFF))
+        return sum
+
+    @staticmethod
+    def byteSum64(val):
```
corthon: What are these? Why are they here?

edk2toollib/acpi/ivrs_parser.py
```diff
@@ -0,0 +1,643 @@
...
+                    self.IVRSBit = 0
+            else:
+                print('Reserved remapping struct found in IVRS table %d' % remapping_header.Type)
+                sys.exit(-1)
+
+            self.ivrs_table.SubStructs.append(remapping_header)
+            # Add to XML
+            self.data = self.data[remapping_header.Length:]
+
+        self.xml = self.toXml()
```
corthon: No need to keep this field updated. Anyone who needs XML format can call EncodeXml().

edk2toollib/acpi/ivrs_parser.py
```diff
@@ -0,0 +1,643 @@
...
+
+    def toXml(self):
```
corthon: For clarity, let's call this ToXmlNode() or ToXmlElementTree() or something like that. ToXml() suggests it would be an XML file or string.

edk2toollib/acpi/ivrs_parser.py
```diff
@@ -0,0 +1,643 @@
...
+        retval = str(self.ivrs_table)
+
+        for sub in self.ivrs_table.SubStructs:
+            retval += str(sub)
+
+        return retval
+
+    def toBytes(self):
+        bytesarray = bytearray(IVRS_TABLE.IVRSHeaderLength)
+        struct.pack_into(IVRS_TABLE.ACPI_TABLE_HEADER.struct_format,
```
corthon: Since this starts at 0 anyway, let's use the more common pack() instead of pack_into().

edk2toollib/acpi/ivrs_parser.py
```diff
@@ -0,0 +1,643 @@
...
+             self.Length,
+             self.Revision,
+             self.Checksum,
+             self.OEMID,
+             self.OEMTableID,
+             self.OEMRevision,
+             self.CreatorID,
+             self.CreatorRevision,
+             self.IVinfo,
+             self.Reserved) = struct.unpack_from(IVRS_TABLE.ACPI_TABLE_HEADER.struct_format, header_byte_array)
```
corthon: Since none of these specify an offset, let's use unpack() rather than unpack_from().

edk2toollib/acpi/ivrs_parser.py
```diff
@@ -0,0 +1,643 @@
...
+                    self.UID =\
+                        struct.unpack("=%ss" % self.UIDLength,
+                                      header_byte_array[IVRS_TABLE.DeviceTableEntryLengthVar:])
+            else:
+                print("Unknown Reserved Device Scope Type Found %d" % self.Type)
+                sys.exit(-1)
+
+            self.data = header_byte_array[:self.Length]
+
+        def set_xml(self, xml_item):
```
corthon: This is the only object that has this pattern. Is there a reason/use-case for it?

edk2toollib/acpi/ivrs_parser.py
```diff
@@ -0,0 +1,643 @@
...
+        # append raw data, update length and checksum
+        self.data += ivmd.data
+        self.ivrs_table.Length += ivmd.Length
+        self.ivrs_table.IVMDlist.append(ivmd)
+        self.ivrs_table.Checksum = self.ivrs_table.calculateACPISum(self.data)
+
+    def IVRSBitEnabled(self):
+        return bool(self.ivrs_table.IVRSBit)
+
+    def CheckIVMDCount(self, goldenxml=None):
```
corthon: Is this just a test function? Why is it a first-class member of the object? Is there a use-case other than testing?

edk2toollib/acpi/ivrs_parser.py
kuqin12: The constants are removed

edk2toollib/acpi/ivrs_parser.py
kuqin12: Encode and Decode methods are added to IVHD, IVMD and DTE classes to replace toBytes and other related parsing methods.

edk2toollib/acpi/ivrs_parser.py
kuqin12: Fixed as suggested.

edk2toollib/acpi/ivrs_parser.py
kuqin12: They disappeared... after update

edk2toollib/acpi/ivrs_parser.py
kuqin12: This is field is not included anymore. The original intention is to keep the consumer side unchanged, but since we are going to modify it anyway, the consumer will be calling EncodeXmlTree when they need xml version of the instance.

edk2toollib/acpi/ivrs_parser.py
kuqin12: This is updated.

edk2toollib/acpi/ivrs_parser.py
kuqin12: This is changed to align with other EncodeXmlNode methods.

edk2toollib/acpi/ivrs_parser.py
kuqin12: This is a test function. It will be moved to one layer up at consumer (audit level).

edk2toollib/acpi/ivrs_parser.py
kuqin12: Updated as suggested.

edk2toollib/acpi/ivrs_parser.py
kuqin12: unpack() needs exact format match (including length), some modifications were made to change them to unpack()

PR #33, opened 2019-10-14
[Bug] multiple class locate in utilityFunctions

matthewfcarlson:
The locate class in module function can now look to see which class to pick in the event that multiple matching classes are found. Preference is given to the module that is in the file of the module that we are loading from. This means if you import a common UefiBuilder, it won't be picked over the UefiBuilder implementation in your settings file.


spbrogan:
I think i need more details of how this logic works.  Meaning can you in function header comments write examples and how you expect this to work?

matthewfcarlson:
Updated to have better comments

PR #39, opened 2019-10-23
base_parser: Fix DSC !ifdef and !ifndef parsing

mdkinney:
The syntax of an !ifdef and !ifndef statement does not
use the value of a define that uses $().  Instead, only
the name of the define is used.

Update the test for !ifdef and !ifndef to check if the
define name in token[1] is in LocalVar or not and push
the correct conditional value of True or False.

This change was tested locally with the edk2/NetworkPkg.
With this change, the defines "Defines" section of
NetworkPkg.ci.yaml can be emptied, and the default values
for all defines are correctly parsed.

Signed-off-by: Kinney <michael.d.kinney@intel.com>

### DIFF
--- a/edk2toollib/uefi/edk2/parsers/base_parser.py
+++ b/edk2toollib/uefi/edk2/parsers/base_parser.py
@@ -274,11 +274,11 @@ class BaseParser(object):
             return True
 
         elif(tokens[0].lower() == "!ifdef"):
-            self.PushConditional((tokens[1].count("$") == 0))
+            self.PushConditional(not (self.LocalVars.get(tokens[1]) is None))
             return True
 
         elif(tokens[0].lower() == "!ifndef"):
-            self.PushConditional((tokens[1].count("$") > 0))
+            self.PushConditional(self.LocalVars.get(tokens[1]) is None)
             return True
 
         elif(tokens[0].lower() == "!else"):
--- a/edk2toollib/uefi/edk2/parsers/base_parser.py
+++ b/edk2toollib/uefi/edk2/parsers/base_parser.py
@@ -274,7 +274,7 @@ class BaseParser(object):
             return True
 
         elif(tokens[0].lower() == "!ifdef"):
-            self.PushConditional(not (self.LocalVars.get(tokens[1]) is None))
+            self.PushConditional(self.LocalVars.get(tokens[1]) is not None)
             return True
 
         elif(tokens[0].lower() == "!ifndef"):


matthewfcarlson:
Looks like it fixes #39

edk2toollib/uefi/edk2/parsers/base_parser.py
```diff
@@ -274,11 +274,11 @@ def ProcessConditional(self, text):
             return True
 
         elif(tokens[0].lower() == "!ifdef"):
-            self.PushConditional((tokens[1].count("$") == 0))
+            self.PushConditional(not (self.LocalVars.get(tokens[1]) is None))
```
matthewfcarlson: I think is not None is a little more pythonic

matthewfcarlson: Though I'm open to keeping it as is if you have strong feelings about it

corthon: I agree with Matt.

mdkinney: I agree.  I will change to "is not None"

mdkinney:
I agree more review is required.  I see the following syntax in OvmfPkg/OvmfPkgIa32.dsc that uses $(<define name>) if an !ifdef statement.  

  #
  # Flash size selection. Setting FD_SIZE_IN_KB on the command line directly to
  # one of the supported values, in place of any of the convenience macros, is
  # permitted.
  #
!ifdef $(FD_SIZE_1MB)
  DEFINE FD_SIZE_IN_KB           = 1024
!else
!ifdef $(FD_SIZE_2MB)
  DEFINE FD_SIZE_IN_KB           = 2048
!else
!ifdef $(FD_SIZE_4MB)
  DEFINE FD_SIZE_IN_KB           = 4096
!else
  DEFINE FD_SIZE_IN_KB           = 4096
!endif
!endif
!endif

The EDK II DSC Specification covers this use case in section 2.2.6.  The $(MACRO) syntax in an !ifdef is not recommended.  It is only supported for backwards compatibility.  Since this use case is still present, the DSC parser in pytool should support it at this time.

https://edk2-docs.gitbooks.io/edk-ii-dsc-specification/content/2_dsc_overview/22_build_description_file_format.html#22-build-description-file-format

I think the improved fix would look at token[1] and strip $() if present and then check the define name.

spbrogan:
Do you have the history of why in this case it is suggested to use without $()?  All other macro usage must have $( ).   

There are two env's need to be checked to see if the macro resolves.  Local should be checked first.
self.LocalVars
self.InputVars

I think maybe a better solution would be to update https://github.com/tianocore/edk2-pytool-library/blob/master/edk2toollib/uefi/edk2/parsers/base_parser.py#L200 as that function is the one that replaces the macros.  

@matthewfcarlson  - Can you add a test case in your PR for both styles of syntax ifdef using $(MACRO) and ifdef MACRO.

spbrogan:
@mdkinney what do you think of this.  #41 .  I want to add one more test for this conditional parsing.  

@matthewfcarlson  - Ignore my first comment to you.  I copied the layout of your unit test and added these.  Enjoy your vacation.  Hopefully not thinking about python. :)

mdkinney:
I believe the reason that $() is not recommended for !ifdef and !ifndef statements is the same reason that $() is not used in a DEFINE statement.  These statements are about the define name, not the value that the define name is assigned.

Another note on this topic is that 2 passes is not sufficient.  Because !ifdef and !ifndef statements can be nested and can appear anywhere in the file, a combination of these statements can active DSC files earlier in the file.  And those newly activated lines can activate other lines earlier in the file.  This means that multiple passes are required and the passes can not stop until there are no new lines activated.

It is critical that the parsers in pytools produce the same results as BaseTools for all support syntax.

spbrogan:
changes have been pushed in to cover the original use case.  There are still parsing inconsistencies but that is something we can work on over time.   
Thanks for reporting this and providing a solution.

PR #41, opened 2019-10-27
Add support for testing special ifdef syntax.  Add unit tests

spbrogan:


edk2toollib/uefi/edk2/parsers/base_parser.py
```diff
@@ -25,6 +25,7 @@ def __init__(self, log):
         self.PPs = []
         self.TargetFile = None
         self.TargetFilePath = None
+        self.MacroNotDefinedValue = "0"  # value to used for undefined macro
```
matthewfcarlson: If we expect this to be somewhat private to the class, maybe we should change this to self._MacroNotDefinedValue

edk2toollib/uefi/edk2/parsers/base_parser.py
```diff
@@ -193,6 +194,21 @@ def PopConditional(self):
...
+
+        v = self.LocalVars.get(token)
+
+        if(v is None):
+            v = self.InputVars.get(token)
+
+        if(v is None):
+            v = self.MacroNotDefinedValue
+
+        if(v.upper() == "TRUE" or v.upper == "FALSE"):
```
matthewfcarlson: v.upper() == "FALSE" instead of v.upper ==

edk2toollib/uefi/edk2/parsers/base_parser.py
```diff
@@ -206,8 +222,18 @@ def ReplaceVariables(self, line):
...
 
         """
-        rep = line.count("$")
+        # first tokenize and look for tokens require special macro
+        # handling without $.  This must be done first otherwise
+        # both syntax options can not be supported.
         result = line
+        tokens = result.split()
+        if tokens[0].lower() in ["!ifdef", "!ifndef"]:
+            if not tokens[1].startswith("$("):
```
matthewfcarlson: We should check that tokens > 2 otherwise we will blow up in strange ways

edk2toollib/uefi/edk2/parsers/base_parser.py
```diff
@@ -274,11 +275,11 @@ def ProcessConditional(self, text):
             return True
 
         elif(tokens[0].lower() == "!ifdef"):
-            self.PushConditional((tokens[1].count("$") == 0))
+            self.PushConditional((tokens[1] != "0"))
```
matthewfcarlson: Should we compare this to the MacroNotDefinedValue?

edk2toollib/uefi/edk2/parsers/base_parser.py
```diff
@@ -274,11 +275,11 @@ def ProcessConditional(self, text):
...
             return True
 
         elif(tokens[0].lower() == "!ifndef"):
-            self.PushConditional((tokens[1].count("$") > 0))
+            self.PushConditional((tokens[1] == "0"))
```
matthewfcarlson: Ditto. Should we compare this to the MacroNotDefinedValue?

edk2toollib/uefi/edk2/parsers/base_parser.py
matthewfcarlson: My guess is that this condition isn't being tested if we can short circuit it

spbrogan:
i agree with the comments.  I'll make an update tomorrow.

matthewfcarlson:
@spbrogan I might respond to some of these comments and push them on

edk2toollib/uefi/edk2/parsers/base_parser.py
matthewfcarlson: Do we want to check if token is TRUE or FALSE, or just look at the replacement value?

PR #42, opened 2019-11-01
Fix a bug where a double Encode() would produce different results.

corthon:
Also start adding unit tests for the Capsule classes.


matthewfcarlson:
@corthon removed the unneeded import. Not sure if you left it because you wanted to implement more tests or not.

corthon:
> @corthon removed the unneeded import. Not sure if you left it because you wanted to implement more tests or not.

Copy-paste error.

corthon:
This is a step towards addressing #43. Still needs more.

PR #50, opened 2019-11-19
Data Object Model for Build objects

matthewfcarlson:
Currently just a start to garner feedback.
Further tests, functionality, and documentation forthcoming


corthon:
Why is file "edk2toollib/uefi/edk2/build_objects/readme.md" in the modified list? Can we fix that?

matthewfcarlson:
That's supposed to be there. I haven't added the content yet

BobCF:
I think the implementation of the dsc parser can re-use the dsc parser in Basetools. And I have a question that if this patch set includes all the interfaces(functions and data structure) that open ci need? I think the dsc parser in Basetools can satisfy these interface in this patch set, for the parser in Basetools provides more functions and more complex data structures.

I am wondering if the parser is moved from basetools to pytools-library, how basetools use those parser?

matthewfcarlson:
I was actually evaluating if it would be easier to simply use the baseparser. It looks like it's a bit of a tangled web of dependencies but it should be doable to extract it to pytools.

I like the architecture of the new parser since it's clean and very modular but I think as we add more features on it is going to get more complicated and less elegant. Plus having two versions of the parser leads to always having some difference between the two. I'd much rather have one than two slightly different implementations.

I think there are a few ways to support the parser scenario for base tools. Let me experiment and see if I can get something we can all be satisfied with

matthewfcarlson:
So to put some notes here, the plan going forward is to create a DSC object that we expect to be populated by some parser. The recipe translator will accept the DSC object and give back a recipe. The recipe will be written out to a DSC that can be used by the build system until we have the ability to invoke the build.py process without a process call.

So the general flow will be: DSC + other files -> DSC translator -> Recipe + transformations -> DSC translator -> composite DSC

For the time being while we wait for the basetools DSC's parser to be integrated in some form, the pytool DSC parser is getting augmented to be somewhat closer to what the basetools version will output. The DSC class will be modeled after the DSC spec rather than the pytool parser.

Some of the newer improvements:
- Proper processing of sections. Previously the DSC parser just wanted to know what modules were in a given DSC rather than what sort of architectures/phases they were in
- Multi line support (you can end a line with / and it will continue onto the new line)
- Build option/Pcd/Sku parsing

BobCF:
I created a Library folder under BaseTools to move the MetaFileParser related file there. It's on my edk2 repo, MetaFileParser branch. https://github.com/BobCF/edk2/tree/MetaFileParser/BaseTools/Source/Python/Library.

And the test script is https://github.com/BobCF/edk2/blob/MetaFileParser/BaseTools/Source/Python/ParserTest.py

The dependency of the MetaFileParser are:
1. Metadata File itself.
2. Workspace and Packages_Path environment variable.

This parser process Macro replacement, !include and other instruction evaluation, and format checking.
The usage and output sample can be found in the test script ParserTest.py.

matthewfcarlson:
This looks fantastic! I'll give this a shot. We're still in a little bit of a tricky spot of how to access the basetools from the pip package but this is excellent progress.

matthewfcarlson:
This PR will eventually be used to get the data objects in, but for now we are using PR #67

matthewfcarlson:
Closing for now

PR #53, opened 2019-11-20
When multiple classes are found by locate_class_in_module, find the one that is the deepest leaf of the desired subclass

### LINKED ISSUE #52: When Multiple settings files are found, pick the one that has the deepest leaf
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
John Schock had a great solution that we'd like to move to

**Describe alternatives you've considered**

**Additional context**

matthewfcarlson:
Fixes #52

### DIFF
--- a/edk2toollib/utility_functions.py
+++ b/edk2toollib/utility_functions.py
@@ -407,7 +407,7 @@ def locate_class_in_module(Module, DesiredClass):
         # class hierarchies, raise an error.
         if _class is not DesiredClass and issubclass(_class, DesiredClass):
             if (DesiredClassInstance is None) or issubclass(_class, DesiredClassInstance):
-              DesiredClassInstance = _class
+                DesiredClassInstance = _class
             elif not issubclass(DesiredClassInstance, _class):
                 raise RuntimeError(f"Multiple instances were found:\n\t{DesiredClassInstance}\n\t{_class}")
     return DesiredClassInstance
--- a/edk2toollib/utility_functions.py
+++ b/edk2toollib/utility_functions.py
@@ -395,27 +395,22 @@ def locate_class_in_module(Module, DesiredClass):
     DesiredClassInstances = []  # a list of all the matching classes that we find
     # Pull out the contents of the module that was provided
     module_contents = dir(Module)
-    module_file = Module.__file__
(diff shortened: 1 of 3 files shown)


edk2toollib/utility_functions.py
```diff
@@ -392,30 +392,25 @@ def locate_class_in_module(Module, DesiredClass):
     there is a class defined in the module that subclasses DesiredClass.
```
corthon: A quick ASCII art here would be useful.

edk2toollib/utility_functions.py
```diff
@@ -392,30 +392,25 @@ def locate_class_in_module(Module, DesiredClass):
...
-    if len(DesiredClassInstances) == 0:  # we didn't find anything
-        return None
-    if len(DesiredClassInstances) > 1:  # we can't log because logging isn't setup yet
-        print(f"Multiple {DesiredClass.__name__} classes were found. Using {DesiredClassInstances[0].__name__}")
-    return DesiredClassInstances[0]  # return the first (highest priority class)
+            if (DesiredClassInstance is None) or issubclass(_class, DesiredClassInstance):
+                DesiredClassInstance = _class
+            elif not issubclass(DesiredClassInstance, _class):
+                raise RuntimeError(f"Multiple instances were found:\n\t{DesiredClassInstance}\n\t{_class}")
+    return DesiredClassInstance
```
corthon: Is there any unit testing around this to ensure that correct classes are returned? This is the sort of thing that might blow up easily if Python were to change its behavior and we'd want to catch that ASAP.

edk2toollib/utility_functions.py
matthewfcarlson: I'll give it a shot

edk2toollib/utility_functions.py
matthewfcarlson: I'll write one

matthewfcarlson: Added a test

edk2toollib/utility_functions.py
matthewfcarlson: Tried to add a quick something

PR #60, opened 2019-11-26
Update readme.md section order

spbrogan:
Because the version history is getting longer the current status is too far down the page.  Update section ordering to move current status up the page

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -33,7 +33,16 @@ NOTE: It is strongly recommended that you use python virtual environments.  Virt
     from edk2toollib.<module> import <class>
     ```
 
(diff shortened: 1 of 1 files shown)


matthewfcarlson:
I wonder if instead we should let the milestones speak to what actually went into the release? At what point do we truncate the version history?

spbrogan:
I think overall we should stop listing changes and defer to milestone and i think we could even just have a link to our milestones and not even describe every release...except when there is something extra interesting.  

Given that we only have 2 milestone i think we should wait a few more or until we are on 0.11.0 or 1.0.0.

PR #61, opened 2019-12-13
[Feature] added locate in MRO and call MRO to utility functions

matthewfcarlson:
This adds some functionality to the utility functions that locates classes in the MRO of a given class as well as a way to call all of those found classes by a given method


matthewfcarlson:
We're going to hold off on this until we know more about the solution we want to take for the diamond inheritance problem.

PR #65, opened 2020-01-03
[Bug] don't do logging in GetHostInfo utility function

matthewfcarlson:
We shouldn't do logging in GetHostInfo as it can be called before logging is fully up and running.


spbrogan:
Does this cause an error or just goes nowhere?

matthewfcarlson:
It causes a logging handler to be created, so you'll get a bunch of garbage in your output. An alternative would be to do our best to clear out logging when we start it up, but that's difficult to make sure you've got everything other than calling shutdown. But then you need to reload it.

PR #69, opened 2020-01-15
[bug] Fixed dec and dsc parser to explode less

matthewfcarlson:
Previously, when the dec or dsc parser encounters an error, it explodes. Previously the API didn't really expect callers to handle exceptions.

### DIFF
--- a/edk2toollib/uefi/edk2/parsers/dec_parser.py
+++ b/edk2toollib/uefi/edk2/parsers/dec_parser.py
@@ -120,7 +120,7 @@ class PcdDeclarationEntry():
         op = sp[2].split("|")
 
         if(len(op) != 4):
-            raise Exception("Too many parts")
+            raise Exception("Too many parts " + rawtext)
 
         self.name = op[0].strip()
         self.default_value = op[1].strip()
@@ -153,6 +153,7 @@ class DecParser(HashFileParser):
         InGuidsSection = False
         InPPISection = False
         InPcdSection = False
+        InExtendedPcd = False
         InIncludesSection = False
 
         for line in self.Lines:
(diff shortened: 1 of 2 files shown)


edk2toollib/uefi/edk2/parsers/dec_parser.py
```diff
@@ -199,6 +200,17 @@ def _Parse(self) -> None:
             elif InPcdSection:
                 if sline.strip()[0] == '[':
                     InPcdSection = False
+                elif sline.strip().endswith("{"):
```
corthon: Is it sufficient to only check the end of the line? Are '{' in the middle of a line a concern? Or has it been preprocessed by this point in the code?

matthewfcarlson:
This will need to be merged in with #71

edk2toollib/uefi/edk2/parsers/dec_parser.py
matthewfcarlson: It will be pre-processed, so comments will be stripped and blank lines skipped

corthon: There's no possibility of an inline struct or something? Like...
```
PcdSomethingSomething|{blah, blah}
```

matthewfcarlson: That's a great point.

matthewfcarlson: Maybe just look at if the count of({) > count of (})

corthon: Dunno. Just trying to poke holes in your work. :-p

matthewfcarlson: No it's good :)

matthewfcarlson:
Is this still needed after #73?

Issue #70, opened 2020-01-16
DEC Parser does not support structured PCD declarations with braces {}


mdkinney:
The DEC parser looks for exactly 4 elements in a PCD declaration statement.

There is an optional 5th element that is an open brace '{' to provide a set of 
additional lines for a structured PCD.  Relax the DEC parser to support up to
5 elements and ignore the DEC file lines between the '{' and '}'

matthewfcarlson:
fixed by #71?

PR #71, opened 2020-01-16
edk2toollib/dec_parser: Add support for structured PCDs

mdkinney:
Update DEC parser to support PCD declarations of a
structured PCD that includes statements between
braces {}.  The information between the braces {} is
skipped.

#70 

Signed-off-by: Michael D Kinney <michael.d.kinney@intel.com>


matthewfcarlson:
Ironically, I have a branch that's nearly identical to this- we even called a few of the variables the same.

matthewfcarlson:
This still needs a test but that has been captured here: #74

PR #79, opened 2020-02-21
[DSC_PARSER] Keep track of all the paths that we find in a DSC

matthewfcarlson:
When parsing a DSC, we should keep track of all the files that are included in a DSC.

### DIFF
--- a/edk2toollib/uefi/edk2/parsers/dsc_parser_test.py
+++ b/edk2toollib/uefi/edk2/parsers/dsc_parser_test.py
@@ -1,0 +1,69 @@
+# @file dsc_parser_test.py
+# Contains unit test routines for the dsc parser class.
+#
+#
+# Copyright (c) Microsoft Corporation
+#
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+import unittest
+import tempfile
+import os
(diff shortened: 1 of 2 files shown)


edk2toollib/uefi/edk2/parsers/dsc_parser_test.py
```diff
@@ -0,0 +1,21 @@
+# @file dsc_parser_test.py
```
corthon: This file seems unrelated...

spbrogan:
In the commit message can you write more details.  We need to start doing better on our commits especially when PR content is not obvious on what it does.

edk2toollib/uefi/edk2/parsers/dsc_parser_test.py
matthewfcarlson: I'm modifying the DSC parser so I think it makes sense to add a test for it :) 
As it stands right now, testing !includes in the parser would be pretty hard without creating a bunch of files. I guess I could do it

matthewfcarlson: Added a much better test

matthewfcarlson:
Commits are trimmed down and fixed up

PR #81, opened 2020-02-25
Adds a library & unit test  for manipulating Windows Firmware Policy structures

out0xb2:
1.  Python library for Windows Firmware Policies that can construct, serialize, deserialize, & print them in a more readable form
2.  Python unittest that parses a basic, known-good policy, then constructs, serializes, & deserializes a policy
3.  Python command-line tool that provides quick interface to deserialize and print binary files, or construct a Windows Firmware Policy binary file from command line arguments.

**NOTE**
All of the above are limited to unsigned policies.  For trust, policies must be signed, where the convention is embed PKCS7.  

More examples and tests to follow.


.flake8
```diff
@@ -1,5 +1,6 @@
 [flake8]
 #E266 too many leading '#' for block comment
 #E722 do not use bare 'except'
-ignore = E266,E722
```
spbrogan: why are you changing this

edk2toollib/windows/policy/firmware_policy_test.py
```diff
@@ -0,0 +1,202 @@
...
+        93 AD 5A FF F8 2B 27 C8 9C 61 85 6C 1A D5 2B B6 1C D5 7A EA 86 5C 08 A0
+        CB 51 21 49 4B F2 04 36 5F 38 48 80 D0 CD FC E5 4D 6A 19 0A 17 86 8D E7
+        4B 32 09 62 15 95 36 61 1C 07 12 B8 D7 64 FA CE 74 84 42 20 18 77 A5 F4
+        4E ED 42 5C 64 BD 57 99 E7 C3 7E 2D 0C 8F FE 38 7B 68 0C 78 DF 3D E9 30
+        10 EA D4 1E 24 E1 60 F1 89 92 59 B4 8C F5 F6 8C F9 CE 8C EB 99 D6 35 0A
+        03 0E FB 88 4A 01 54 8C 20 84 18 F5 49 6B 8B 96 D3 55 45 0C 99 79 23 28
+        51 18 C2 4B 96 02 74 79 86 44 FB B5 98 59 69 E2 80 DE""")
+
+    def test_basic_decode(self):
+        inp = io.BytesIO(FirmwarePolicyTest.TEST_DEVICEID_C2E28)
```
spbrogan: maybe you should specify what C2E28 is and then compare policyManuf.<fields here> with their actual values.

edk2toollib/windows/policy/firmware_policy_tool.py
```diff
@@ -0,0 +1,73 @@
+# @file
```
spbrogan: i don't think this tool belongs here.  CLI / tools go into pytool extensions

edk2toollib/windows/policy/firmware_policy.py
```diff
@@ -0,0 +1,563 @@
...
+##
+# The RULE structure is as follows
+##
+# UINT32 RootKey
+# UINT32 OffsetToSubKeyName  # ValueTable offset
+# UINT32 OffsetToValueName   # ValueTable offset
+# UINT32 OffsetToValue       # ValueTable offset to a PolicyValueType + PolicyValue
+#
+##
+# Each Rule corresponds to 3 entries in the ValueTable
```
spbrogan: should add tests for the things below.  Especially interesting is the reuse of subkeys

edk2toollib/windows/policy/firmware_policy.py
```diff
@@ -0,0 +1,563 @@
...
+##
+
+import io
+import struct
+import uuid
+
+STRICT = False
+
+
+# Class for storing, serializing, deserializing, & printing RULE elements
```
spbrogan: should move to docstrings for class and function comments.

edk2toollib/windows/policy/firmware_policy.py
```diff
@@ -0,0 +1,563 @@
...
+                 OffsetToSubKeyName=0, OffsetToValueName=0, OffsetToValue=0):
+        self.RootKey = RootKey
+        self.OffsetToSubKeyName = OffsetToSubKeyName
+        self.OffsetToValueName = OffsetToValueName
+        self.OffsetToValue = OffsetToValue
+        self.SubKeyName = SubKeyName
+        self.ValueName = ValueName
+        self.Value = Value
+
+    def __eq__(self, other):
```
spbrogan: should try to use newer python syntax for parameter types and return types

edk2toollib/windows/policy/firmware_policy.py
```diff
@@ -0,0 +1,563 @@
...
+    #
+    # 16-bit, little endian, size of the string in _bytes_
+    # followed by a UTF-16LE string, no NULL terminator
+    #
+    # Example of "PlatformID"
+    # \x14\x00P\x00l\x00a\x00t\x00f\x00o\x00r\x00m\x00I\x00D\x00
+    #
+    # The trailing \x00 is not a NULL, it is UTF-16LE encoding of "D"
+
+    def __init__(self, fs=None, offset=None, String=None):
```
spbrogan: why offset.  It looks like it is optional but you never seek back....so why not make the caller seek to the right point to start and the caller fix the fs pointer on return.

spbrogan:
also please write up a basic docs in md format describing this.
https://github.com/tianocore/edk2-pytool-library/tree/master/docs/features

.flake8
out0xb2: Adding W503 to the ignore list.  W503 requires that multi-line conditionals have the boolean dangling on the prior line where you have to go find it.  The new, preferred syntax is to have the booleans at the beginning of the continued line.  

See lines 82 plus or minus of firmware_policy.py (I'll add a comment) which W503 did not like where I put the boolean operators at the beginning of continued lines, not the end of prior lines.  I agree with Knuth, PEP8, and the community at large who said that this prior anti-pattern should actually be the recommended pattern (i.e. W503 should be ignored, consider requiring W504).

PEP8:  https://github.com/python/peps/commit/c59c4376ad233a62ca4b3a6060c81368bd21e85b
https://github.com/PyCQA/pycodestyle/issues/513
https://github.com/python-diamond/Diamond/commit/b62bc3fe91a138613c464846f1650797c1df6f9c

edk2toollib/windows/policy/firmware_policy.py
out0xb2: firmware_policy.py serialize functions do **not** currently perform this optimization, but some other libraries generate them.  The FromFileStream() methods _have_ been tested to work against several optimized policies.

- In your opinion, should we add this optimization as an option to the FirmwarePolicy Serialize() method?
- if yes, then we can build the test cases using this library.  If no, then I'll locate some binary blobs created by other libraries.

edk2toollib/windows/policy/firmware_policy.py
```diff
@@ -0,0 +1,563 @@
...
+        self.SubKeyName = SubKeyName
+        self.ValueName = ValueName
+        self.Value = Value
+
+    def __eq__(self, other):
+        if (self.RootKey == other.RootKey
+            # Rule table offset are not considered for equivalency
+            and self.SubKeyName == other.SubKeyName
+            and self.ValueName == other.ValueName
+                and self.Value == other.Value):
```
out0xb2: The lines above are where Flake8 W503, I prefer the new style (that violates W503) where the binary operator begins the continued line

edk2toollib/windows/policy/firmware_policy.py
out0xb2: It pushes the seek down into the worker classes so that the high level code is more readable IMHO.
A 3 line seek/call/restore can become a 1 line call with offset parameter.  Because the value table is not guaranteed sequential, every value requires 3 seeks.

out0xb2:
> looks pretty good.
> You might want to look over your code coverage to add a few simple test cases that will raise your coverage and strengthen your validation.

I took a look, coverage is at 81% as-is:
Name                                                                  Stmts   Miss  Cover   Missing
---------------------------------------------------------------------------------------------------
firmware_policy.py                                                      344     66    81%   84, 93, 162-165, 175, 219, 241-244, 252, 258, 262-264, 268, 277, 289, 293-295, 307-316, 319-321, 325-328, 331-339, 395, 427-430, 448-453, 475, 484, 488-490, 499, 502-504, 515, 520, 532, 549, 556

edk2toollib/windows/policy/firmware_policy.py
spbrogan: no need to add optimization....but maybe a unit test with existing policy just to make sure that functionality doesn't break.

edk2toollib/windows/policy/firmware_policy.py
spbrogan: so does the caller do a seek after return to bring FS back to expected location? 

Is it fair to say fs on entry should always point to head of value table?  Should FS just be a value table parameter instead of file stream.  Meaning should the outer object parse the value table as a bye array and then pass that to the inner objects?  This way you would also contain your seek length within the value table

edk2toollib/windows/policy/firmware_policy.py
spbrogan: ok.  might be nice to figure out how we check in a file for auto formatting to respect our flake8 settings.  I think you had issue here and on line length.

edk2toollib/windows/policy/firmware_policy_tool.py
```diff
@@ -0,0 +1,73 @@
...
+    try:
+        with open(filename, "rb") as f:
+            policy = FirmwarePolicy(fs=f)
+            policy.Print()
+
+    except FileNotFoundError:
+        print('ERROR:  File not found: "{0}"'.format(filename))
+
+
+def CreatePolicyFromParameters(filename, manf, product, sn, nonce, oem1, oem2):
```
matthewfcarlson: Maybe print instead of create? We don't return the policy we create

edk2toollib/windows/policy/firmware_policy.py
out0xb2: Do you mean https://docs.python.org/3/library/typing.html , or something else?

edk2toollib/windows/policy/firmware_policy.py
out0xb2: Line length fixed, I finally found a way to break them that satisfied me.

The autoformatter, autopep8 by default, seems to enforce neither rule, perhaps due to the controversy.  So you could end up with the boolean expression before or after the line break when you "autoformat".  I'm sure there's a config to fix that, but time... tick tock

edk2toollib/windows/policy/firmware_policy_tool.py
out0xb2: This is the CLI tool that Sean says I should move to extensions.
CreatePolicyFromParameters() initializes a Firmware Policy object with the parameters, opens a file "filename", and the Serialize(f) method writes it to file.  Print is here as a CLI debugging convenience.  My takeaway is that I should consider a better function name and add some documentation.  Noted, thanks!

edk2toollib/windows/policy/firmware_policy.py
out0xb2: The upcoming commit will add a test case and a large, optimized policy to be tested.

edk2toollib/windows/policy/firmware_policy.py
out0xb2: autopep8 should use .flake8 ... if in the same directory?  I'm closing this for now, if merited I'd suggest you file an issue tracker feature request?

edk2toollib/windows/policy/firmware_policy_test.py
out0xb2: Upcoming commit will do this and more

edk2toollib/windows/policy/firmware_policy_tool.py
out0xb2: For now, I am going to remove this tool, and use this feedback when I PR it into Pytools extensions

edk2toollib/windows/policy/firmware_policy_test.py
out0xb2: Addressed in commit 2a966083e3b43c04a5f6c6ab338108bcdfe9244c

edk2toollib/windows/policy/firmware_policy_tool.py
out0xb2: Removed in 2a966083e3b43c04a5f6c6ab338108bcdfe9244c

edk2toollib/windows/policy/firmware_policy.py
out0xb2: Committed in e89edefe55dac2101829619ffcc1c2b1626ab2e8

edk2toollib/windows/policy/firmware_policy.py
out0xb2: Committed in e89edefe55dac2101829619ffcc1c2b1626ab2e8

out0xb2:
> > looks pretty good.
> > You might want to look over your code coverage to add a few simple test cases that will raise your coverage and strengthen your validation.

Code coverage with additional tests is at 91%

edk2toollib/windows/policy/firmware_policy.py
spbrogan: yes.  i generally only do simple types that don't require more imports like str, int, etc.

edk2toollib/windows/policy/firmware_policy.py
```diff
@@ -0,0 +1,561 @@
+# @file
+#
+# Library to parse/manage/build unsigned firmware policy blobs
+#
+# Copyright (c), Microsoft Corporation
```
corthon: Inconsistent copyright formatting.

edk2toollib/windows/policy/firmware_policy.py
```diff
@@ -0,0 +1,561 @@
...
+# For example:
+#   RootKey = 0xEF100000
+#   SubKeyName = "PolicyGroupFoo"
+#   ValueName = "TpmClear"
+#   MyExampleKey = EF100000\PolicyGroupFoo\TpmClear
+#
+# Each value has a type that can be primitive or a variable-length structure.
+#
+##
+# The policy binary, pack(1), little endian
```
corthon: Can you link to where each of these structures is defined in C?

edk2toollib/windows/policy/firmware_policy.py
```diff
@@ -0,0 +1,589 @@
...
+        if fs is None:
+            self.FormatVersion = self.POLICY_FORMAT_VERSION
+            self.PolicyVersion = self.POLICY_VERSION
+            self.PolicyPublisher = self.POLICY_PUBLISHER
+            self.Reserved1Count = 0
+            self.Reserved1 = []
+            self.OptionFlags = 0
+            self.Reserved2Count = 0
+            self.RulesCount = 0
+            self.Reserved2 = []
```
corthon: This is confusing. This is a list of Reserved2 objects, but Reserved1 is a list of... something else?

edk2toollib/windows/policy/firmware_policy.py
out0xb2: Fix pending for next commit.
**NOTE:** I copied/pasted from MU_BASECORE\BaseTools\Edk2ToolsBuild.py ... [line 86](https://github.com/microsoft/mu_basecore/blob/d7307756c69c6da1a7aaa554441bf431d2694501/BaseTools/Edk2ToolsBuild.py#L86).  Shall I fix that too?

edk2toollib/windows/policy/firmware_policy.py
out0xb2: You're looking at the egg, the 🐔 will hatch in another repo soon.  At that time, great point.  Where do we keep track of TODOs, Issue Tracker?

edk2toollib/windows/policy/firmware_policy.py
out0xb2: There are 2 arrays of different types that are not used in the Firmware Policy, Reserved1[] and Reserved2[].  When the Reserved1Count and Reserved2Count are both equal to 0, then the arrays are not present in the structure.  Cannot eliminate the Counts as they will be expected in the structure server-side.

edk2toollib/windows/policy/firmware_policy.py
out0xb2: Completed in latest commit

edk2toollib/windows/policy/firmware_policy.py
out0xb2: Fixed in latest commit

edk2toollib/windows/policy/firmware_policy.py
out0xb2: I took a note to self to add this later when the C struct is posted

edk2toollib/windows/policy/firmware_policy.py
out0xb2: I have 2 unit tests for Reserved2.  The first one uses an old, non firmware policy that included Reserved2 entries.  The second one is synthetic, exercising some unused methods to ensure that they are basically functioning for future use.  I'm closing this for now, but re-open if you have strong concern here.  Still working on some other issues.

out0xb2:
> also please write up a basic docs in md format describing this.
> https://github.com/tianocore/edk2-pytool-library/tree/master/docs/features

@spbrogan Complete as of latest commit

edk2toollib/windows/policy/firmware_policy.py
out0xb2: addressed in the 10th commit 29d5833845f3e9c31652f9b740786fdef65285df

edk2toollib/windows/policy/firmware_policy.py
```diff
@@ -0,0 +1,641 @@
...
+                raise Exception(ErrorMessage)
+
+        self.vt = Type
+
+    """if offset is not specified, stream fs position is at beginning PolicyValueType struct"""
+    @classmethod
+    def FromFileStream(cls, fs: BinaryIO, fsOffset: int = None):
+        if fsOffset:
+            fs.seek(fsOffset)
+        valueType = struct.unpack('<H', fs.read(2))[0]
```
corthon: Can this leverage the FromBytes() method once the file has been read?

edk2toollib/windows/policy/firmware_policy.py
```diff
@@ -0,0 +1,641 @@
...
+        return cls(valueType)
+
+    @classmethod
+    def FromBytes(cls, b: bytes, bOffset: int = 0):
+        valueType = struct.unpack_from('<H', b, bOffset)[0]
```
corthon: Please use the self.StructFormat if this is truly the same pack string.

edk2toollib/windows/policy/firmware_policy.py
```diff
@@ -0,0 +1,641 @@
...
+        if (len(LocalString) != (StringLength / 2)):
+            raise Exception('String length mismatch')
+        return cls(String=LocalString)
+
+    def Print(self, prefix: str = ''):
+        print('%s%s%s' % (prefix, 'String:  ', self.String))
+
+    def Serialize(self, valueOut: bytearray):
+        b = str.encode(self.String, encoding='utf_16_le')
+        size = struct.pack('<H', len(b))
```
corthon: Is this the length? self.StringLengthFormat ?

edk2toollib/windows/policy/firmware_policy.py
out0xb2: Good question, you've helped point out where I could make this more understandable.

This method cannot do as you ask because `fsOffset` is the offset from the beginning of the file, where `baOffset` in `FromBytes()` is the offset from the beginning of the ValueTable.

The file can be deserialized sequentially until it encounters the "ValueTable", which requires random access to process.  Streams were originally used, and when the ValueTable was encountered, `seek() `was invoked numerous times to perform the random accesss.
  
This commit 10 incarnation deserializes using streams until it encounters the ValueTable, then to avoid the seeks(), the ValueTable is consumed in 1 read() into Bytes, and further parsing has been switched to use FromBytes().

What I'm inferring is that I might consider reading the whole file into Bytes to begin with, and do all of the parsing on that, instead of stream half the type, bytes half the time?

edk2toollib/windows/policy/firmware_policy.py
out0xb2: Good catch, thanks!

edk2toollib/windows/policy/firmware_policy.py
out0xb2: You caught me again, refactored part of the code but dropped the other... will fix asap, thanks!

out0xb2: Addressed in latest commit

edk2toollib/windows/policy/firmware_policy.py
out0xb2: Addressed in recent commit

matthewfcarlson:
Should we change the title since it no longer includes a CLI tool?

out0xb2:
We can spell, @corthon please approve

PR #85, opened 2020-02-29
[Ivrs parser] fixes all the spelling errors to allow PR's through

matthewfcarlson:


.cspell.json
```diff
@@ -76,6 +76,9 @@
         "hexlify",
         "ctypes",
         "ioapic",
-        "OEMID"
+        "OEMID",
+        "iommu",
+        "IOMMUEFR",
+        "IVMD"
```
out0xb2: IOMMU is generic, but are IOMMUEFR and IVMD AMD-specific acronyms?  If yes, then perhaps they should be local to the ivrs_parser*.py files not global in .cspell.json?

out0xb2:
@corthon can you help please?

.cspell.json
matthewfcarlson: So I have two choices if I want to keep them inside the ivrs_parser. I could change the name, or I could put comments that disable the spell checker. There's only one word file list. So if we want them to be valid words, this is the only spot they can go.

matthewfcarlson:
The current way cspell is structured, there isn't a mechanism to have separate dictionaries other than just the global one. Though I can think of a few solutions that could add that functionality. However, it would be a change in basecore to the cspell tool itself.

.cspell.json
out0xb2: Since there are only 2 AMD-IOMMU-specific files, I'd lean toward the per-file syntax:
# spell-checker:ignore IVMD, IOMMUEFR

But just my opinion 😁

matthewfcarlson: Oh that's an option? Let's do that.

matthewfcarlson: An excellent suggestion

out0xb2:
@corthon , please help approve

PR #87, opened 2020-02-29
Add returncode to standard RunCmd output

### LINKED ISSUE #86: RunCmd should print the returncode to ease debugging
Printing the returncode of the called command would ease debugging.

singweeMsftGitHub:
Pull request intended to fix #86.

### DIFF
--- a/edk2toollib/utility_functions.py
+++ b/edk2toollib/utility_functions.py
@@ -266,6 +266,7 @@ def DetachedSignWithSignTool(SignToolPath, ToSignFilePath, SignatureOutputFile, 
     params = params + ' /debug /v "' + ToSignFilePath + '" '
     ret = RunCmd(SignToolPath, params)
     if(ret != 0):
+        logging.error("Signtool failed %d" % ret)
         return ret
     signedfile = os.path.join(OutputDir, os.path.basename(ToSignFilePath) + ".p7")
     if(not os.path.isfile(signedfile)):
@@ -302,7 +303,10 @@ def CatalogSignWithSignTool(SignToolPath, ToSignFilePath, PfxFilePath, PfxPass=N
         # add password if set
         params = params + ' /p ' + PfxPass
     params = params + ' /debug /v "' + ToSignFilePath + '" '
-    return RunCmd(SignToolPath, params, workingdir=OutputDir)
+    ret = RunCmd(SignToolPath, params, workingdir=OutputDir)
+    if(ret != 0):
+        logging.error("Signtool failed %d" % ret)
+    return ret
 
 
 ###
--- a/edk2toollib/utility_functions.py
+++ b/edk2toollib/utility_functions.py
@@ -185,7 +185,7 @@ def RunCmd(cmd, parameters, capture=True, workingdir=None, outfile=None, outstre
     logging.log(logging_level, "------------------------------------------------")
 
     if raise_exception_on_nonzero and c.returncode != 0:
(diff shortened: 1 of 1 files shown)


edk2toollib/utility_functions.py
```diff
@@ -177,9 +177,11 @@ def RunCmd(cmd, parameters, capture=True, workingdir=None, outfile=None, outstre
     endtime = datetime.datetime.now()
     delta = endtime - starttime
     endtime_str = "{0[0]:02}:{0[1]:02}".format(divmod(delta.seconds, 60))
+    returncode_str = "{0:#010x}".format(c.returncode)
```
matthewfcarlson: You can also just do returncode_str = hex(c.returncode) I believe.

matthewfcarlson: Either way I think is fine

edk2toollib/utility_functions.py
```diff
@@ -177,9 +177,11 @@ def RunCmd(cmd, parameters, capture=True, workingdir=None, outfile=None, outstre
...
     logging.log(logging_level, "------------------------------------------------")
     logging.log(logging_level, "--------------Cmd Output Finished---------------")
     logging.log(logging_level, "--------- Running Time (mm:ss): " + endtime_str + " ----------")
+    logging.log(logging_level, "----------- Return Code: " + returncode_str + " ------------")
```
spbrogan: Line 188 - the other place returncode is printed should be updated to match...or maybe not even needed.

edk2toollib/utility_functions.py
singweeMsftGitHub: Thanks for looking at this.  The reason I chose .format() was to limit the size of the converted returncode so that the dashes would all line up nicely in predictable fashion (I assumed a 32 bit returncode).

edk2toollib/utility_functions.py
singweeMsftGitHub: Thanks for the feedback.  @spbrogan, I've removed the other locations where returncode is printed.
For line 188, where we raise an exception, do you have a preference of whether I leave the exception string empty or change the string?  For the time being, I've changed the string.  But I'm open to making this whatever you'd like it to be.

spbrogan: looks fine for now.  I don't have a strong opinion.

edk2toollib/utility_functions.py
```diff
@@ -264,7 +266,6 @@ def DetachedSignWithSignTool(SignToolPath, ToSignFilePath, SignatureOutputFile,
     params = params + ' /debug /v "' + ToSignFilePath + '" '
     ret = RunCmd(SignToolPath, params)
     if(ret != 0):
-        logging.error("Signtool failed %d" % ret)
         return ret
```
spbrogan: this change will lead to different results because the return code will be printed at info level instead of error.  Probably would prefer to leave this here.

singweeMsftGitHub: Got it.  I'll undo the changes on line 267 and also around line 307 as well, since that's the same thing (I had removed a logging.error print).

singweeMsftGitHub: Got it.  I've gone ahead and reverted this portion of the change, along with a similar change around line 305.

edk2toollib/utility_functions.py
singweeMsftGitHub: Thanks!

PR #90, opened 2020-03-21
Change Enum to Python 3

matthewfcarlson:
Python 3.4+ includes support for real enums. 
This class is offered as a compatibility shim.

### DIFF
--- a/edk2toollib/utility_functions_test.py
+++ b/edk2toollib/utility_functions_test.py
@@ -65,5 +65,32 @@ class UtilityFunctionsTest(unittest.TestCase):
         self.assertEqual(found_class, GrandChildOfDesiredClass)
 
 
+class EnumClassTest(unittest.TestCase):
+    def test_EnumClass_array(self):
+        all_animals = ["Dog", "Cat", "Rabbit"]
+        animals = utilities.Enum(all_animals)
+        
+        # make sure we have three animals
+        self.assertEqual(len(animals), 3)
+        self.assertTrue(hasattr(animals, "Dog"))
+        self.assertFalse(hasattr(animals, "Rat"))
+        self.assertFalse(hasattr(animals, "dog"))
+        # check to make sure the values are unique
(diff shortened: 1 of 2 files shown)


spbrogan:
since we know the users can we just convert them over to EnumInt or the std Enum.  I am not sure the feature is even used anymore anyway.

matthewfcarlson:
I just added it for the sake of backwards compatibility

edk2toollib/utility_functions_test.py
```diff
@@ -65,5 +65,32 @@ def test_locate_class_in_module(self):
...
+        self.assertFalse(hasattr(animals, "dog"))
+        # check to make sure the values are unique
+        self.assertNotEqual(animals.Dog, animals.Cat)
+
+        # make sure we can iterate over the members
+        for animal in animals.__members__:
+            self.assertIn(animal, all_animals)
+
+
+    def test_EnumClass_args(self):
```
corthon: Can we add a test for "not in enum"?

corthon:
I agree with the "let's move over to the native thing". Is there some way we could keep this back-compat PR, but spit out a warning to get people to migrate so that we can drop this in the next major version? @spbrogan, would that even be desirable?

edk2toollib/utility_functions_test.py
matthewfcarlson: We do have a test to make sure it's not in.
``` python
self.assertFalse(hasattr(animals, "Rat"))
self.assertFalse(hasattr(animals, "dog"))
```

We could create two seperate enums and check to make sure they aren't in each other

matthewfcarlson:
I think that would be pretty easy to do, since we have our own constructor. I like Bret's solution.

edk2toollib/utility_functions_test.py
corthon: Didn't notice that.

PR #92, opened 2020-04-02
Add more conditional to the unittest that have caused the parser to fail

spbrogan:
Add unit tests that mimic the failing cases found in Edk2 EmulatorPkg.dsc


matthewfcarlson:
Replaced by !94

matthewfcarlson:
#94

PR #96, opened 2020-04-14
Added coverage file to report code coverage

matthewfcarlson:


matthewfcarlson:
This may or may not work

matthewfcarlson:
https://github.com/MicrosoftDocs/codecoverage-yaml-samples
https://docs.microsoft.com/en-us/azure/devops/pipelines/test/codecoverage-for-pullrequests?view=azure-devops#coverage-status-details-and-indicators

spbrogan:
in the docs there are a lot of limitations.  Azure repos and .coverage files only.  We don't do either of those so do you think this will work?

matthewfcarlson:
This doesn't work. We'll need to go with something codecov.io

PR #100, opened 2020-04-30
Path Utilities refactor and unit tests.

spbrogan:


spbrogan:
- [ ] Add unit tests showing the failure condition
- [ ] Add unit tests that break more assumptions
- [ ] Eval other functions in the same module

edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -56,25 +56,31 @@ def __init__(self, ws, packagepathlist):
...
         relpath = None
         found = False
         if abspath is None:
             return None
-        for a in self.PackagePathList:
-            stripped = abspath.lower().partition(a.lower())[2]
-            if stripped:
-                # found our path...now lets correct for case
+        for a in (os.path.normcase(p) for p in self.PackagePathList):
+            if os.path.normcase(abspath).startswith(a):
```
matthewfcarlson, line 70: wouldn't it make more sense to convert abspath into normabspath or something rather than doing it each time?

edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -116,34 +131,56 @@ def GetAbsolutePathOnThisSytemFromEdk2RelativePath(self, relpath):
...
+                self.logger.info("PackagePath is: %s" % os.pathsep.join(self.PackagePathList))
+                self.logger.info("Workspace path is : %s" % self.WorkspacePath)
+                return None
+
+        # InputPath is in workspace or PackagesPath for worst case scenario.
+
         dirpathprevious = os.path.dirname(InputPath)
         dirpath = os.path.dirname(InputPath)
-        while(dirpath is not None):
+        while(True):
```
matthewfcarlson, line 156: I would like some other termination condition

edk2toollib/uefi/edk2/path_utilities_test.py
```diff
@@ -0,0 +1,575 @@
...
+import os
+import sys
+import tempfile
+import shutil
+from edk2toollib.uefi.edk2.path_utilities import Edk2Path
+
+
+class PathUtilitiesTest(unittest.TestCase):
+
+    def MakeFile(self, parentpath, filename):
```
matthewfcarlson, line 20: MakeFile is a little generic. Maybe a little mores specific or a docstring?

edk2toollib/uefi/edk2/path_utilities_test.py
```diff
@@ -0,0 +1,575 @@
...
+        # absolute path
+        with self.assertRaises(Exception):
+            Edk2Path(self.tmp, [pp_full_1])
+        # relative path
+        with self.assertRaises(Exception):
+            Edk2Path(self.tmp, [pp])
+
+    def test_pp_inside_workspace(self):
+        ''' test with packagespath pointing to folder nested inside workspace
+        root/                   <-- current working directory
```
matthewfcarlson, line 100: These comments are fantastic!

matthewfcarlson:
I wish I could see the improvement in code coverage :/

edk2toollib/uefi/edk2/path_utilities.py
spbrogan: this isn't in a loop and will only do the conversion on the fly so it we break it won't even convert all of them.  I thought about doing this at the class level but it didn't seem right and it added complexity for no practical gain.  

If you look at GetContainingPackage i did convert them into a local list.  This is because it is called in a loop.

edk2toollib/uefi/edk2/path_utilities.py
spbrogan: I agree but i couldn't find a condition and during unit test I could not execute failed path.  If you look at code coverage you will that is the only code not executed.  

I guess i could have a counter to avoid an infinite loop.  Something like 200.   Do you like that better?  I still won't be able to execute it as i don't think it is possible with path length limits.

edk2toollib/uefi/edk2/path_utilities_test.py
spbrogan: I'll make private and add docstring.

edk2toollib/uefi/edk2/path_utilities.py
spbrogan: @corthon and @matthewfcarlson  waiting on ideas if this is worth it?

matthewfcarlson: I'm fine with just a counter to avoid infinite loops

spbrogan: please confirm you are ok with solution.

PR #103, opened 2020-05-12
[Bug] Fix Build report parser enum warning and convert INF path to Edk2 style

spbrogan:


matthewfcarlson:
This looks like it just does the first thing, not the second

edk2toollib/uefi/edk2/parsers/buildreport_parser.py
```diff
@@ -140,7 +140,7 @@ def Parse(self):
                                 logging.debug("Parsing Mod: %s" % value)
                                 self.Name = value
                             elif(key == "module inf path"):
-                                self.InfPath = value
+                                self.InfPath = value.replace("\\", "/")
```
spbrogan, line 143: @matthewfcarlson  this is the change needed for edk2 INF  path

matthewfcarlson: Looks good to me

PR #109, opened 2020-06-10
Added custom targets to RunCmd

matthewfcarlson:


corthon:
I think this may be too many fixes in a single PR.

edk2toollib/utility_functions.py
```diff
@@ -54,17 +54,21 @@ def Enum(*args):
...
             else:
                 self.ret = self._target(*self._args, **self._kwargs)
+        except SystemExit as e:
+            self.ret = e.code
         except BaseException as e:
             self.exc = e
 
-    def join(self, timeout=None):
-        super(PropagatingThread, self).join()
+    def join(self, timeout=0.5):
```
corthon, line 69: Is this timeout change to address its own issue?

matthewfcarlson:
This is a nasty change but would allow some really important behaviors:
1. We've had a request to detect when edk2 build is doing a long running task and output some message. This would allow that.
2. In the case of an OVMF buiild, it would allow the platformbuilder to watch the output from OVMF for asserts and automatically kill QEMU. This would be highly useful for CI and headless environments

edk2toollib/utility_functions_test.py
```diff
@@ -91,5 +149,10 @@ def test_EnumClass_args(self):
         # check to make sure the values are unique
         self.assertNotEqual(colors.Green, colors.Red)
 
+
 # DO NOT PUT A MAIN FUNCTION HERE
 # this test runs itself to test runpython script, which is a tad bit strange yes.
+test = UtilityFunctionsTest()
```
matthewfcarlson, line 155: This shouldn't have gotten pushed up

corthon, line 155: Adding these tests would seem to be in contradiction to the above note, no?

matthewfcarlson: Yes- pytest on my dev box hangs so I run the tests manually. This is me being dumb and pushing up without checking carefully

edk2toollib/utility_functions.py
matthewfcarlson: It's related- since we call join with a specific timeout in RunCmd, if we didn't pass it through it wouldn't work

spbrogan:
I think we need to find a less messy and impactful way to do this.  I don't have the answer yet.

matthewfcarlson:
Agreed- this is a next release kind of problem

PR #110, opened 2020-06-12
No fail mode for DSC parser

matthewfcarlson:


edk2toollib/uefi/edk2/parsers/dsc_parser.py
```diff
@@ -237,16 +238,35 @@ def ParseInfPathMod(self, line):
...
+                        raise
+                    else:
+                        self.Logger.warning(e)
 
     def __ProcessDefines(self, lines):
         if(len(lines) > 0):
             for l in lines:
-                (line, add) = self.__ParseDefineLine(l)
-                self.__ProcessDefines(add)
+                try:
```
spbrogan, line 255: why is this needed?

matthewfcarlson: Because we do include processing here. I was trying to catch the lines where we processed includes

spbrogan: why do we do it in two spots?

matthewfcarlson: Because the DSC parser is a two pass system

matthewfcarlson: It processes it once to find defines (including in includes), then does the proper parse. This is still a hold-over from the limited parser that was pre project mu

matthewfcarlson:
@spbrogan how is this for comments?

spbrogan:
Tested and works with OvmfPkg in Mu before and after dependencies have been populated.

PR #114, opened 2020-06-20
Relative Includes on DSCs

### LINKED ISSUE #105: stuart_ci_build does not support execution from paths other than the workspace root
stuart_ci_build is working when I execute it from the workspace root, but fails when I run it from a subdirectory.

For example, the following works from the Mu_Plus repo workspace root:
```mu_plus>stuart_ci_build -c .pytool\CISettings.py -p FirmwarePolicyPkg```

But when I change to a subdirectory:
```mu_plus\FirmwarePolicyPkg>stuart_ci_build -c ..\.pytool\CISettings.py -p FirmwarePolicyPkg```

it fails with
```
ERROR - --->Test Failed: Compiler Plugin DEBUG returned 1
PROGRESS - --Running FirmwarePolicyPkg: Compiler Plugin RELEASE --
CRITICAL - EXCEPTION: [Errno 2] No such file or directory: 'FirmwarePolicyPkg/FirmwarePolicyPkg.dsc'
CRITICAL - Traceback (most recent call last):
   File "c:\git\venv_1911\lib\site-packages\edk2toolext\invocables\edk2_ci_build.py", line 226, in Go
    tc, plugin_output_stream)
   File "C:\git\FwPolicy\fw_policy\MU_BASECORE\.pytool\Plugin\CompilerPlugin\CompilerPlugin.py", line 80, in RunBuildPlugin
    dp.ParseFile(AP_Path)
   File "c:\git\venv_1911\lib\site-packages\edk2toollib\uefi\edk2\parsers\dsc_parser.py", line 259, in ParseFile
    f = open(os.path.join(filepath), "r")
 FileNotFoundError: [Errno 2] No such file or directory: 'FirmwarePolicyPkg/FirmwarePolicyPkg.dsc'

ERROR - --->Test Failed: Compiler Plugin RELEASE returned 1
```

### PR DESCRIPTION
**Pull request by matthewfcarlson**
Fixes #105

### DIFF
--- a/edk2toollib/uefi/edk2/parsers/base_parser.py
+++ b/edk2toollib/uefi/edk2/parsers/base_parser.py
@@ -26,8 +26,8 @@ class BaseParser(object):
         self.RootPath = ""
         self.PPs = []
         self._Edk2PathUtil = None
-        self.TargetFile = None
-        self.TargetFilePath = None
+        self.TargetFile = None # the relative path of the target file
+        self.TargetFilePath = None # the abs path of the target file
         self.CurrentLine = -1
         self._MacroNotDefinedValue = "0"  # value to used for undefined macro
 
@@ -81,39 +81,37 @@ class BaseParser(object):
 
     def FindPath(self, *p):
         """
-        Given a path, it will find it relative to the root, the current target file, or the 
+        Given a path, it will find it relative to the root, the current target file, or the packages path
         Args:
           *p:
 
-        Returns:
+        Returns: None on failure
 
         """
-        # NOTE: Some of this logic should be replaced
-        #       with the path resolution from Edk2Module code.
-
+        print(f"Finding {p}")
         # If the absolute path exists, return it.
         Path = os.path.join(self.RootPath, *p)
         if os.path.exists(Path):
-            return Path
+            return os.path.abspath(Path)
 
         # If that fails, check a path relative to the target file.
         if self.TargetFilePath is not None:
             Path = os.path.abspath(os.path.join(os.path.dirname(self.TargetFilePath), *p))
             print("Checking " + Path + " " + self.TargetFilePath)
             if os.path.exists(Path):
-                return Path
+                return os.path.abspath(Path)
 
         # If that fails, check in every possible Pkg path.
         if self._Edk2PathUtil is not None:
             target_path = os.path.join(*p)
             Path = self._Edk2PathUtil.GetAbsolutePathOnThisSytemFromEdk2RelativePath(target_path)
             if Path is not None:
-                return Path
+                return os.path.abspath(Path)
 
         # log invalid file path
         Path = os.path.join(self.RootPath, *p)
         self.Logger.error("Invalid file path %s" % Path)
-        return Path
+        return None
 
     def WriteLinesToFile(self, filepath):
         """
--- a/edk2toollib/uefi/edk2/parsers/base_parser.py
+++ b/edk2toollib/uefi/edk2/parsers/base_parser.py
@@ -44,7 +44,7 @@ class BaseParser(object):
         Returns:
 
         """
-        self.RootPath = path
+        self.RootPath = os.path.abspath(path)
         self._ConfigEdk2PathUtil()
         return self
 
@@ -85,7 +85,7 @@ class BaseParser(object):
         Args:
           *p:
 
-        Returns: None on failure
+        Returns: a full absolute path if the file exists, None on failure
 
         """
         # If the absolute path exists, return it.
@@ -104,11 +104,11 @@ class BaseParser(object):
             target_path = os.path.join(*p)
             Path = self._Edk2PathUtil.GetAbsolutePathOnThisSytemFromEdk2RelativePath(target_path, False)
             if Path is not None:
-                return os.path.abspath(Path)
+                return Path
 
         # log invalid file path
         Path = os.path.join(self.RootPath, *p)
-        self.Logger.error("Invalid file path %s" % Path)
+        self.Logger.error(f"Invalid file path: {p}")
         return None
 
     def WriteLinesToFile(self, filepath):
--- a/edk2toollib/uefi/edk2/parsers/base_parser.py
+++ b/edk2toollib/uefi/edk2/parsers/base_parser.py
@@ -102,7 +102,7 @@ class BaseParser(object):
         # If that fails, check in every possible Pkg path.
         if self._Edk2PathUtil is not None:
             target_path = os.path.join(*p)
-            Path = self._Edk2PathUtil.GetAbsolutePathOnThisSytemFromEdk2RelativePath(target_path)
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: edk2toollib/uefi/edk2/parsers/base_parser.py [hunk 1]
```diff
@@ -71,37 +81,35 @@ def SetInputVars(self, inputdict):
...
 
         # If that fails, check in every possible Pkg path.
-        for Pkg in self.PPs:
-            Path = os.path.join(self.RootPath, Pkg, *p)
-            if os.path.exists(Path):
-                return Path
+        if self._Edk2PathUtil is not None:
+            target_path = os.path.join(*p)
+            Path = self._Edk2PathUtil.GetAbsolutePathOnThisSytemFromEdk2RelativePath(target_path, False)
+            if Path is not None:
```
spbrogan, line 106: Isn't Path already absolute?  Given the function name GetAbsolutePathOnThisSytemFromEdk2RelativePath?

edk2toollib/uefi/edk2/parsers/base_parser.py
```diff
@@ -71,37 +81,35 @@ def SetInputVars(self, inputdict):
...
+                return os.path.abspath(Path)
 
         # log invalid file path
         Path = os.path.join(self.RootPath, *p)
         self.Logger.error("Invalid file path %s" % Path)
```
spbrogan, line 111: If logging an error i wouldn't join the path as that seems wrong.   I would suggest the error be something like the the *p path is not valid.

edk2toollib/uefi/edk2/parsers/base_parser.py
```diff
@@ -71,37 +81,35 @@ def SetInputVars(self, inputdict):
...
           *p:
 
-        Returns:
+        Returns: None on failure
 
         """
-        # NOTE: Some of this logic should be replaced
-        #       with the path resolution from Edk2Module code.
-
         # If the absolute path exists, return it.
```
spbrogan, line 91: you never do this?  You never treat it as absolute.  If that is supported you should check it and then not do any path modifications if it is absolute. Also seems unit test should have caught this if it was intended to work

edk2toollib/uefi/edk2/parsers/base_parser.py
```diff
@@ -71,37 +81,35 @@ def SetInputVars(self, inputdict):
 
     def FindPath(self, *p):
         """
-
+        Given a path, it will find it relative to the root, the current target file, or the packages path
         Args:
           *p:
 
-        Returns:
+        Returns: None on failure
```
spbrogan, line 88: on success it is is returning a full absolute path to a file in the local file system

edk2toollib/uefi/edk2/parsers/base_parser.py
```diff
@@ -24,8 +25,9 @@ def __init__(self, log="BaseParser"):
         self.ConditionalStack = []
         self.RootPath = ""
```
spbrogan, line 26: Is this Workspace?  Is this abstracted from Workspace on purpose?

edk2toollib/uefi/edk2/parsers/base_parser.py
```diff
@@ -24,8 +25,9 @@ def __init__(self, log="BaseParser"):
...
         self.PPs = []
-        self.TargetFile = None
-        self.TargetFilePath = None
+        self._Edk2PathUtil = None
+        self.TargetFile = None  # the relative path of the target file
```
spbrogan, line 29: relative to what?

edk2toollib/uefi/edk2/parsers/base_parser.py
```diff
@@ -43,18 +45,26 @@ def SetBaseAbsPath(self, path):
 
         """
         self.RootPath = path
```
spbrogan, line 47: what is path?  absolute or relative.  Relative to what (CWD?)

edk2toollib/uefi/edk2/parsers/base_parser.py
matthewfcarlson: Fair

edk2toollib/uefi/edk2/parsers/base_parser.py
matthewfcarlson: Done

edk2toollib/uefi/edk2/parsers/base_parser.py
matthewfcarlson: This doesn't seem to make any sense?

edk2toollib/uefi/edk2/parsers/base_parser.py
matthewfcarlson: updated comment

edk2toollib/uefi/edk2/parsers/base_parser.py
matthewfcarlson: self.RootPath? Yes it's effectively RootPath.

edk2toollib/uefi/edk2/parsers/base_parser.py
matthewfcarlson: I'm not super clear on what TargetFile and TargetFilePath do. They were already there.

edk2toollib/uefi/edk2/parsers/base_parser.py
matthewfcarlson: alright- converted to an abs path

edk2toollib/uefi/edk2/parsers/base_parser.py
spbrogan: ok.  we need to figure it out.

edk2toollib/uefi/edk2/parsers/base_parser.py
spbrogan: I would have expected the code to if os.path.isabs(p) and os.path.exists(p) 

The code never checks if it is absolute.  It always joins it with rootpath which would not work if the incoming path was something like c:\MyFolder\MyFile.dsc  

This would turn out to be something c:\MyFolder\c:\MyFolder\MyFile.dsc

Does that make sense?

matthewfcarlson: So the code you have highlighted isn't clear what you're commenting on

edk2toollib/uefi/edk2/parsers/base_parser.py
matthewfcarlson: Let me see if we can drop targetfile

matthewfcarlson:
So this has one major flaw. It we have a DSC with two includes and they're both relative. It will fail since our target include path fails.

edk2toollib/uefi/edk2/parsers/base_parser.py
matthewfcarlson: Dropped

matthewfcarlson:
@spbrogan it's ready

PR #117, opened 2020-09-08
Fixing DMAR parser break caused by classes name change

kuqin12:
This change fixes "AttributeError: 'DMARTable' object has no attribute '**'" due to class name change inside DMAR table class.

The issue was introduced during porting process from Project Mu.


matthewfcarlson:
Do you have a commit where this broke? Just for reference

matthewfcarlson (approved):

matthewfcarlson:
I'm happy to merge this once you're happy Kun

kuqin12:
> I'm happy to merge this once you're happy Kun

Let's merge it. I tried to poke for ANDD and ASTR example, but there is nothing readily available. We can add it once there is one. Thanks.

PR #125, opened 2021-04-02
Add CPER Parsers

TaylorBeebe:


corthon:
We generally avoid putting any CLI tools in the "library". The tool component should move to "extensions".

However, the classes for manipulating the data can come out into modules that live here. That should make the tool pretty small.

corthon (commented):
I don't think it makes sense to call this PR (and the associated files) parsing tools any more because they aren't tools. They're just parsers.

edk2toollib/windows/telem/cper_parser_tool.py
```diff
@@ -0,0 +1,1001 @@
...
+                    "e71254e8-c1b9-4940-ab76-909703a4320f": "Generic Section",
+                    "e71254e9-c1b9-4940-ab76-909703a4320f": "WHEA Error Packet Section",
+                    }
+
+# List of plugin classes which inherit from CPER_SECTION_DATA and are
+# therefore capable of parsing section data
+Parsers = SECTION_DATA_PARSER.__subclasses__()
+
+
+class GENERIC_FIELD(object):
```
corthon, line 131: Class names should be CamelCase, not UPPER_SNAKE_CASE.

edk2toollib/windows/telem/cper_parser_tool.py
```diff
@@ -0,0 +1,1001 @@
...
+                    "6f3380d1-6eb0-497f-a578-4d4c65a71617": "IPF SAL Record Section",
+                    "8a1e1d01-42f9-4557-9c33-565e5cc3f7e8": "XPF MCA Section",
+                    "e71254e7-c1b9-4940-ab76-909703a4320f": "NMI Section",
+                    "e71254e8-c1b9-4940-ab76-909703a4320f": "Generic Section",
+                    "e71254e9-c1b9-4940-ab76-909703a4320f": "WHEA Error Packet Section",
+                    }
+
+# List of plugin classes which inherit from CPER_SECTION_DATA and are
+# therefore capable of parsing section data
+Parsers = SECTION_DATA_PARSER.__subclasses__()
```
corthon, line 128: Meanwhile, globals should be UPPER_SNAKE_CASE, not CamelCase.

edk2toollib/windows/telem/cper_parser_tool.py
```diff
@@ -0,0 +1,1001 @@
...
+
+
+class GENERIC_FIELD(object):
+    '''A super class for fields of CPER records'''
+
+    def __init__(self, raw_value):
+        self.raw_value = raw_value
```
corthon, line 135: This pattern doesn't exist anywhere else in the encoding object we have in edk2toollib (to my knowledge). The object themselves should just be shells that contain the accessible fields. Then you should call `decode()` on either an `IOStream` or a `bytes` object (Sean and I disagree on this, but either approach is fine) to populate the contents. Then, ideally, there should also be an `encode()` method that would put the object back into either `bytes` or an `IOStream` such that `MyObject.decode(my_bytes).encode() == my_bytes`. But just having the `decode()` method is fine for a parser.

edk2toollib/windows/telem/cper_parser_tool.py
```diff
@@ -0,0 +1,1001 @@
...
+        self.parsed_value = ""
+        self.Parse()
+
+    def Parse(self):
```
corthon, line 139: Please change all the `Parse` functions to `Decode` or `decode` for consistency and taking in the bytes to decode (rather than coming from a private field).

edk2toollib/windows/telem/cper_parser_tool.py
```diff
@@ -0,0 +1,1001 @@
...
+        self.Parse(valid_bits_field)
+
+    def Parse(self, valid_bits_field):
+        if(valid_bits_field.FruStringValid()):
+            self.parsed_value = "UNIMPLEMENTED"
+        else:
+            self.parsed_value = "Invalid"
+
+
+class CPER(object):
```
corthon, line 537: This class has too many lines that don't have test coverage.

edk2toollib/windows/telem/cper_parser_tool.py
```diff
@@ -0,0 +1,1001 @@
...
+
+    def GetSectionsFruStringList(self) -> list:
+        "Returns the fru string of each section as a list"
+        temp = []
+        for sec in self.section_headers:
+            temp.append(sec.GetFruString(False))
+        return temp
+
+
+class CPER_HEADER(object):
```
corthon, line 749: This class has too many lines without test coverage.

edk2toollib/windows/telem/cper_parser_tool.py
```diff
@@ -0,0 +1,1001 @@
...
+        if not Friendly:
+            temp = AttemptPrettyUuid(self.GetCreatorId())
+        else:
+            temp = str(self.GetCreatorId(not Friendly))
+        string += "Creator Id:     " + temp + '\n'
+
+        return string 
+
+
+class CPER_SECTION_HEADER(object):
```
corthon, line 882: This class has too many lines without test coverage.

corthon (changes requested):
There's some quick find/replace stuff here, and a good bit more code coverage needed. Will evaluate the dynamic parsing stuff next.

PR #127, opened 2021-04-16
Modify utility_functions reader() implementation to be more permissive

joschock:
Modify utility_functions reader() implementation to be more permissive in the face of unexpected encoding coming from utilities executed with RunCmd()

### DIFF
--- a/edk2toollib/utility_functions.py
+++ b/edk2toollib/utility_functions.py
@@ -162,6 +162,8 @@ def timing(f):
 # @param logging_level - log level to log output at.  Default is INFO
 # @param raise_exception_on_nonzero - Setting to true causes exception to be raised if the cmd
 #                                     return code is not zero.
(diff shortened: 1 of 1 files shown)


spbrogan (changes requested):
How about making it an optional requirement to runcmd.  Default would be error as it is now.   But this would allow a caller to ignore the errors if so desired.

joschock:
> How about making it an optional requirement to runcmd. Default would be error as it is now. But this would allow a caller to ignore the errors if so desired.

Done.

edk2toollib/utility_functions.py
```diff
@@ -166,7 +166,7 @@ def wrap(*args):
 # @return returncode of called cmd
```
spbrogan, line 166: Need a function header comment for the variable...but otherwise looks good.

joschock: Thanks, added.

corthon (approved):

PR #128, opened 2021-04-30
Helper classes for variable attributes and policies

corthon:


edk2toollib/uefi/edk2/variable_policy.py
```diff
@@ -0,0 +1,171 @@
...
+        self.Name = None
+
+    def __str__(self):
+        return "VARIABLE_LOCK_ON_VAR_STATE_POLICY(%s, %d, %s)" % (self.Namespace, self.Value, self.Name)
+
+    def decode(self, buffer):
+        """
+        load this object from a bytes buffer
+
+        return any remaining buffer
```
spbrogan, line 37: not as implemented

edk2toollib/uefi/edk2/variable_policy.py
```diff
@@ -0,0 +1,171 @@
...
+        """
+        (_namespace, self.Value, _) = struct.unpack(
+            self._HdrStructFormat, buffer[:self._HdrStructSize])
+
+        self.Namespace = uuid.UUID(bytes_le=_namespace)
+
+        if len(buffer) > self._HdrStructSize:
+            self.Name = buffer[self._HdrStructSize:].decode(
```
spbrogan, line 45: should you check type before you blindly try to decode?
Should you also figure out size rather than just taking everything and making it a string

edk2toollib/uefi/edk2/variable_policy.py
```diff
@@ -0,0 +1,171 @@
...
+                'utf-16').strip('\x00')
+
+        return b''
```
spbrogan, line 48: This is not what function header says it will do

spbrogan (changes requested):
Needs more unit tests for the variable policy stuff

edk2toollib/uefi/edk2/variable_policy.py
corthon: Fixed.

edk2toollib/uefi/edk2/variable_policy.py
corthon: Fixed.
Also, I'm going to let the decode just blow up if it's not a proper utf16 string.

edk2toollib/uefi/edk2/variable_policy.py
corthon: It is now, buddy boy.

spbrogan (approved):

edk2toollib/uefi/edk2/variable_policy.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def decode(self, buffer):
+        """
+        load this object from a bytes buffer
+
+        return any remaining buffer
+        """
+        (self.Version, self.Size, self.OffsetToName, _namespace,
+            self.MinSize, self.MaxSize, self.AttributesMustHave,
+            self.AttributesCantHave, self.LockPolicyType, _) = struct.unpack(
```
spbrogan, line 171: I think you should check the revision so that you fail gracefully rather than have to debug
ENTRY_REVISION

spbrogan (changes requested):

edk2toollib/uefi/edk2/variable_policy.py
```diff
@@ -0,0 +1,183 @@
...
+                       self.LockPolicy.Name, self.LockPolicy.Value]
+
+        result += ["0x%08X" % self.MinSize,
+                   "0x%08X" % self.MaxSize,
+                   str(EfiVariableAttributes(self.AttributesMustHave)),
+                   str(EfiVariableAttributes(self.AttributesCantHave))]
+
+        return result
+
+    def decode(self, buffer):
```
spbrogan, line 163: adding some error handling and at least failing with sanity when things don't add up will make it much easier to root cause issues found in an input blob.  Version, Name, LockPolicyType should all have some sort of basic ability to fail gracefully.

corthon: Fixed all but Name, which I don't understand.

spbrogan (approved):

spbrogan:
Thanks.

PR #132, opened 2021-07-01
Resolve the unittest error in the initial 0.10.14 release

corthon:
A new unittest in #130 compared a fixed value to a date, which worked when submitted for PR, but fails on every other day (until the UNIX epoch wraps). This update corrects that and updates the dynamic value.


corthon:
No idea how we're going to version this for a release "fix".

spbrogan (approved):

PR #133, opened 2021-07-19
Fix `SECTION GUIDED` parsing

brandonaut:
`sectionValue` wasn't being used and it broke the parsing on sections that didn't have a `x = y` part of the section.

### DIFF
--- a/edk2toollib/uefi/edk2/parsers/fdf_parser.py
+++ b/edk2toollib/uefi/edk2/parsers/fdf_parser.py
@@ -143,7 +143,6 @@ class FdfParser(HashFileParser):
                             if sline.upper().startswith("SECTION GUIDED"):  # get the guided section
                                 section_def = sline[14:].strip().split("=", 1)
                                 sectionType = section_def[0].strip()  # UI in this example
-                                sectionValue = section_def[1].strip()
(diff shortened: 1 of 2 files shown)


spbrogan:
Could you add a unit test that shows what breaks and then provide the fix?

After looking it seems we have zero unit tests for the FDF parser.  Thats because these parsers were really designed to be thrown away once edk2-basetools was cleaned up so the modules could be used without invoking edk2build.  

I can start the unit test or if you are willing.  I was just going to follow the pattern like: https://github.com/tianocore/edk2-pytool-library/blob/master/edk2toollib/uefi/edk2/parsers/dec_parser_test.py

brandonaut:
There are a few existing tests in https://github.com/tianocore/edk2-pytool-library/blob/master/edk2toollib/uefi/edk2/parsers/test/fdf_parser_test.py, but they only test the DEFINES. 

I'll look at adding some unit tests. Probably won't get to it this week though.

brandonaut:
@corthon @spbrogan Sorry for the delay. I've added a basic unit test, but I confess I don't understand the FDF spec well enough to know how to test the SECTION GUIDED properly. The spec seems at odds with what the parser expects, but I am probably just not understanding the spec.

spbrogan (approved):

PR #134, opened 2021-07-29
Update encoding to use locale based detection when parsing logs

kuqin12:
Update encoding to use locale based detection when parsing logs.

### DIFF
--- a/edk2toollib/utility_functions.py
+++ b/edk2toollib/utility_functions.py
@@ -20,7 +20,7 @@ import platform
 import importlib
 from collections import namedtuple
 from enum import Enum as StdEnum
-
+import locale
 
 def Enum(*args):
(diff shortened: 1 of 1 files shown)


corthon:
What sort of testing was done to verify that:
1) This works as advertised and
2) Doesn't break anything?

kuqin12:
> What sort of testing was done to verify that:
> 
> 1. This works as advertised and
> 2. Doesn't break anything?

This was tested on both En-US and Zh-CN Windows machines. The texts output from cl is properly relayed to final output file. But the impact on gcc and/or linux is unknown to this point. Could you point me to where those scenarios could be tested? I can run the same change there as well.

spbrogan (approved):

PR #136, opened 2021-08-12
Add RemoveTree function to consolidate 

spbrogan:


edk2toollib/utility_functions.py
```diff
@@ -434,12 +434,29 @@ def locate_class_in_module(Module, DesiredClass):
...
+    '''
+
+    def remove_readonly(func, path, _):
+        ''' private function to attempt to change permissions on file/folder being deleted'''
+        os.chmod(path, stat.S_IWRITE)
+        func(path)
+
+    for _ in range(3):  # retry up to 3 times
+        try:
+            shutil.rmtree(dir_path, ignore_errors=ignore_errors, onerror=remove_readonly)
```
joschock, line 456: does ignore_errors=ignore_errors still throw OSErrror on remove failure? If not, it may not be what you want since it could leave parts of the directory intact (which might cause other more subtle failures if caller expects it to be  fully removed).

joschock: ah, nevermind; I see what you are doing there.

joschock (approved):

Issue #101, opened 2020-05-01
path utilities GetAbsolutePathOnThisSytemFromEdk2RelativePath is spelled wrong


spbrogan:
should be GetAbsolutePathOnThisSystemFromEdk2RelativePath

There are a few callers in the extensions.  So need to figure out right way to fix this.

PR #137, opened 2021-08-19
Fix incorrectly spelled function.

spbrogan:


spbrogan:
resolves #101

spbrogan:
closes #101

PR #139, opened 2021-09-10
Add workarounds for build report parsing

edwinzMSFT:


spbrogan:
can you add unit tests for both of these situations.  That will help identify the syntax that causes the error and keep from regression later.

edwinzMSFT:
> can you add unit tests for both of these situations. That will help identify the syntax that causes the error and keep from regression later.

Added a unit test for the change in path_utilities.py, however it seems like there isn't test coverage for the build report parser as a whole right now. Would I need to implement it all from scratch?

spbrogan (approved):

spbrogan:
I am breaking this into two PRs.  
I made minor modification to your change for path parsing to be more complete (not depend on "_")
Will close this PR without merging because it will be merged in other PRs.

#140 
#142

PR #140, opened 2021-09-17
Fix relative path resolution when packages path list contains multiple paths which are substrings of each other.

spbrogan:


spbrogan:
Broke out one of the changes from Edwin's PR https://github.com/tianocore/edk2-pytool-library/pull/139
Credited Edwin with this change since I only authored a minor tweak.  
Thanks

PR #151, opened 2022-03-04
Stop buildreport_parser.py from attempting to search for binary compo…

edwinzMSFT:
…nents in module list.

### DIFF
--- a/edk2toollib/uefi/edk2/parsers/buildreport_parser.py
+++ b/edk2toollib/uefi/edk2/parsers/buildreport_parser.py
@@ -290,7 +290,7 @@ class BuildReport(object):
 
                 # Take absolute path and convert to EDK build path
                 RelativePath = self.PathConverter.GetEdk2RelativePathFromAbsolutePath(i)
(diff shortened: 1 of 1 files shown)


spbrogan (changes requested):
lets find a better and more sustainable way to fix the path issues.

Javagedes:
@edwinzMSFT is this PR in regard to the sporadic failing of build_report described in #152 ? If so, I've explained the issue there. I've submitted #153 that will catch the incorrect package path at build rather then causing a failure down the line in build_report.

edwinzMSFT:
Nested package path check is a better way of fixing this: https://github.com/tianocore/edk2-pytool-library/pull/153

edwinzMSFT:
> @edwinzMSFT is this PR in regard to the sporadic failing of build_report described in #152 ? If so, I've explained the issue there. I've submitted #153 that will catch the incorrect package path at build rather then causing a failure down the line in build_report.

Yep, I've went ahead and closed this one.

PR #153, opened 2022-04-08
Add nested package paths check

Javagedes:
Add nested package paths check to ensure no package path exists inside of another package path. This causes issues with path utilities (particularly GetEdk2RelativePathFromAbsolutePath())

### DIFF
--- a/azure-pipelines/windows-test-pr-ci.yml
+++ b/azure-pipelines/windows-test-pr-ci.yml
@@ -22,7 +22,7 @@ pr:
 jobs:
 - template: azure-pipelines/templates/build-test-job.yml@pytool_extensions
   parameters:
-    vm_image: 'windows-latest'
+    vm_image: 'windows-2019'
     root_package_folder: "edk2toollib"
     name: 'windows'
     codecov_enabled: true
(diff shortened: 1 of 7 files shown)


spbrogan:
before completing this we should check specs and think about if a layout as described below is allowed/reasonable.

WS Root
  - Pkg1
  - Pkg2
  - Folder1
    - Pkg3
    - Pkg4
    - Folder2
      - Pkg5
      - Pkg6
  
In this case Folder1 and Folder2 would be on the PackagePath and there would still not be any nested "packages".

spbrogan:
My suggestion is to create a couple of good unit tests first.  See what "valid" scenarios fail.  Then we can work them to pass and/or document what is "valid".  

I agree that PackagesPath order is important, but I also agree that a bias towards finding the relative path based on the package root (if it exists) should be used.  But beware not all files are in a package.

Javagedes:
@spbrogan Ready for review. This update passes all old tests and two new tests. Old tests include when a file is not in a workspace, which was a concern of yours. The new updates continue to properly return "None". 

Additionally, I might make a few updates to GetContainingPackage() in a separate PR. I think what I did at the end of __init__(),  might be a more elegant solution for preventing infinite loops, than the current solution of `for _ in range(100)` Let me know if you think it's worth while!

spbrogan (approved):

Issue #154, opened 2022-04-11
Upgrade windows-test-pr-ci vm_image to windows-2022


Javagedes:
windows-test-pr-ci.yml had to be downgraded to windows-2019 as windows-2022 does not have a valid Windows 11 WDK, which includes necessary executables such as inf2cat.exe.

Once Windows-2022 (https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md) contains the Windows 11 WDK, windows-test-pr-ci.yml vm_image can be upgraded to windows-2022 and this PR can be closed. See the below PR's for this issue in the github repo:

https://github.com/actions/virtual-environments/issues/5386
https://github.com/actions/virtual-environments/issues/5235

Javagedes:
Additionally, azure-pipelines-release.yml had to be downgraded to windows-2019 until Windows 11 WDK is released.

Javagedes:
The WDK, for a short amount of time, was added to the windows-2022 vm. It has since been reverted due to issues.

https://github.com/actions/virtual-environments/commit/5690ae918350fa3a27805b6397e9e2e7acc035c8

Javagedes:
Looks like WDK was added back in:

https://github.com/actions/virtual-environments/pull/5838

PR #155, opened 2022-04-11
Downgrade pipeline to windows-2019

Javagedes:
windows-2022 does not have WDK installed due to there not being an official release for windows 11. WDK is necessary to pass ci, so downgrading to windows-2019 until there is a proper WDK for windows-2022.


spbrogan (approved):

spbrogan:
This PR will need to be reverted once the VM Image is improved. 
See tracking bug here: https://github.com/tianocore/edk2-pytool-library/issues/154

PR #156, opened 2022-04-22
Add long-line handling to build report parser for INF paths

zurcher:
The Build Report will line break after 120 characters which causes parser errors when Module INF Paths are longer than 101 characters. This change reads additional lines until the Module INF Path contains ".inf".


zurcher:
Filed issue #157

Javagedes (approved):

Javagedes:
When a Module INF Path wraps to a new line within a Module Summary section, the parser cuts off the INF path at the new line, resulting in an incorrect parsed value for the INF path. Due to this, FindComponentByInfPath() fails and the fully parsed module will be missing the FV that it is a part of. This results in a non-halting error but does mean that the data the parser is providing is incorrect.

PR #159, opened 2022-05-25
Introduce close_fds argument in RunCmd

jgarver:
This is necessary to enable GNU make's job control to pass through
commands like stuart_build.  GNU make's job control is implemented using
tokens on file descriptor.  If the file descriptors remain open, a GNU
make calling stuart_build can communicate with a GNU make called by
stuart_build.  However, Python's Popen closes file descriptors by
default.

This change adds a close_fds argument to RunCmd, which passes through to
Popen.  It defaults to True, which is Popen's default value, so existing
behaviors are not changed.

A similar change will be proposed in edk2-pytool-extensions.

### DIFF
--- a/edk2toollib/utility_functions.py
+++ b/edk2toollib/utility_functions.py
@@ -167,11 +167,14 @@ def timing(f):
 #                                     return code is not zero.
 # @param encodingErrors - may be given to set the desired error handling for encoding errors decoding cmd output.
 #                         Default is 'strict'.
+# @param close_fds - If True, file descriptors are closed before the command is run.
+#                    Default is True.
 #
 # @return returncode of called cmd
 ####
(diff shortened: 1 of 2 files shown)


Javagedes:
@jgarver I've resolved the Cspell / NodeJS issue, so you should be good there.

Can you raise an issue for this PR and a provide a reproduceable situation where this error arrises so that I can take a look? Being as the default value does not change, this should be no issue, but I'd like to fully understand the issue and reproduce it before I approve the PR.

Thanks!

jgarver:
Added #165 for this.

Also, reminder that for stuart to actually benefit from this change, we'll need to make matching changes in edk2-pytool-extensions.  If you think you'll accept this change, I'll go ahead and post the edk2-pytool-extensions change.  This change will need to go in first, though.

Thanks!

spbrogan (approved):

PR #160, opened 2022-06-14
Bugfix - Standardizing separators for packages paths

bkarstens:
This makes the paths returned by `GetAbsolutePathOnThisSystemFromEdk2RelativePath` have consistent separators. Previously, if `self.PackagePathList` had `/`s as separators, the returned path would have both `/` and `\` as separators. This can cause issues with other tools.

### DIFF
--- a/edk2toollib/uefi/edk2/path_utilities.py
+++ b/edk2toollib/uefi/edk2/path_utilities.py
@@ -108,16 +108,16 @@ class Edk2Path(object):
 
         if relpath is None:
             return None
-        relpath = Path(relpath)
-        abspath = Path(self.WorkspacePath) / relpath
-        if abspath.exists():
-            return str(abspath)
+        relpath = relpath.replace("/", os.sep).replace('\\', os.sep)
+        abspath = os.path.join(self.WorkspacePath, relpath)
+        if os.path.exists(abspath):
+            return abspath
 
         for package_path in self.PackagePathList:
-            package_path = Path(package_path)
-            abspath = package_path / relpath
-            if abspath.exists():
-                return str(abspath)
+            package_path = package_path.replace("/", os.sep).replace('\\', os.sep)
+            abspath = os.path.join(package_path, relpath)
+            if(os.path.exists(abspath)):
+                return abspath
         if log_errors:
             self.logger.error("Failed to convert Edk2Relative Path to an Absolute Path on this system.")
             self.logger.error("Relative Path: %s" % relpath)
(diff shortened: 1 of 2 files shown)


spbrogan:
can you author a test case that shows the error.

bkarstens:
While working on this, I ran into the unpleasant realization that `\` is a valid character in file paths on Linux. So if someone was crazy enough to do that, they would run into some issues with my potential fix.

spbrogan:
Going to move this to release 11.6 just in case it breaks something not covered. :)

edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -117,19 +119,19 @@ def GetAbsolutePathOnThisSystemFromEdk2RelativePath(self, relpath, log_errors=Tr
 
         if relpath is None:
             return None
-        relpath = relpath.replace("/", os.sep)
+        relpath = relpath.replace("/", os.sep).replace('\\', os.sep)
```
spbrogan, line 122: can you add comments throughout this module (where necessary) that indicates functions will return paths using the os.sep.  

There are probably some functions that should return edk2 paths and those should be careful to always return paths with "/".

spbrogan, line 122: instead of the replace logic should you just use https://docs.python.org/3/library/os.path.html#os.path.normpath

should this be used on all input strings that are going to be filesystem specific?

bkarstens: That works on Windows, but not on Linux 😢. Linux thinks `\` is a valid character for file and folder names, so it assumes you have a folder or file named `foo\bar` and won't clean them up:
```console
root@tryit-active:~# mkdir foo                                                                                                                         
root@tryit-active:~# mkdir foo\\bar                                                                                                                    
root@tryit-active:~# ls -l                                                                                                                             
total 0                                                                                                                                                
drwxr-xr-x 1 root root 0 Jul 14 02:14  foo                                                                                                             
drwxr-xr-x 1 root root 0 Jul 14 02:15 'foo\bar' 
```

If some functions are expected to return an edk2 path that uses `/` regardless of OS, I don't have a great solution. Returning os.sep from everything *except sometimes `/`* isn't very user-friendly. Forcing `/` as the separator in all cases might cause a bunch of trouble for users? Or we could not make a guarantee and say the bug is in the tool that was expecting the separators to all be the same (the one I ran into was the guid check plugin).

Javagedes:
@bkarstens I had a couple PR's that were holding up your PR. They are now in. One of which caused a conflict with your PR, however the two will be simple to resolve as they are just flake8 formatting conflicts. Once you have those fixed and the comments resolved, I'll get this in so that we can do a release.

Javagedes:
Hi @bkarstens We've made many changes to path_utilities. Can you check again and see if this PR is still valid, and if so, update it accordingly? Thanks

Javagedes:
closing as stale. Feel free to re-open.

PR #169, opened 2022-08-01
Add GetEdk2RelativePathFromAbsolutePath() unit test for poor path names

Javagedes:
Add the ability handle path names where the package name shows up in the package path before the package path. As an example, would be` ClientPkg/CLIENT_PKG/ClientPkg/ClientPkg.dec`

### DIFF
--- a/edk2toollib/uefi/edk2/path_utilities_test.py
+++ b/edk2toollib/uefi/edk2/path_utilities_test.py
@@ -823,7 +823,11 @@ class PathUtilitiesTest(unittest.TestCase):
 
         # pass in bad parameter
         self.assertIsNone(pathobj.GetAbsolutePathOnThisSystemFromEdk2RelativePath(None))
+<<<<<<< HEAD
 
+=======
+    
+>>>>>>> bc9f31b (Initial update)
     def test_get_absolute_path_then_relative_path_when_path_contains_repeated_packagepath_name(self):
         ''' Test the back and forth between GetAbsolutePath and GetRelativeFromAbsolute when the
         path structure has multiple instances of a package path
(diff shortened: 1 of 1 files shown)


edk2toollib/uefi/edk2/path_utilities_test.py
```diff
@@ -761,3 +761,44 @@ def test_get_absolute_path_on_this_system_from_edk2_relative_path(self):
...
+        ws_rel = "PlatformClientPkg"
+        ws_abs = os.path.join(self.tmp, ws_rel)
+        os.mkdir(ws_abs)
+
+        folder_pp1_rel = "CLIENT_PKG"
+        folder_pp1_abs = os.path.join(ws_abs, folder_pp1_rel)
+        os.mkdir(folder_pp1_abs)
+
+        ws_pkg_name = "ClientPkg"
+        ws_pkg_abs = self._make_edk2_package_helper(folder_pp1_abs, ws_pkg_name)
```
spbrogan, line 790: I don't see you actually create the ClientPkg or ClientPkg.dec file or the Module1/2 stuff?  Comment should match code and test case.  Am I missing something?

edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -97,7 +97,11 @@ def GetEdk2RelativePathFromAbsolutePath(self, abspath):
 
         package = self.GetContainingPackage(abspath)
         if package is not None:
-            relpath = abspath[str(temp_path).find(package):]
+            if f'{package.upper()}.DEC' in str(temp_path).upper():
```
spbrogan, line 100: maybe i am dense but can you write out in comments what you are trying to account for here.   I think I might need comments written out for the previous change too which included the temp_path stuff above and slicing based on temp_path.

spbrogan (changes requested):
need some additional clean up, code comments for context about the different conditionals, test comment clean up.  

Overall, appreciate the submission of test case with code change.

edk2toollib/uefi/edk2/path_utilities_test.py
Javagedes: Module1 / Module2 is a remanent from copy / pasting from another test. Will remove from the comment as it is not important.

As for creating ClientPkg / ClientPkg.dec, The files don't actually need to exist to be able to determine the relative path. This capability / functionality is used for PR evaluation when PR has a file or directory removed. However, I think it is best that the test still creates ClientPkg and ClientPkg.dec, so I will add that in.

cfernald (approved):

spbrogan (approved):

PR #172, opened 2022-08-03
PR evaluation CI failure bugfix

Javagedes:
Current PR evaluation CI fails in edk2-pytool-extension for three separate tests due to recent changes in edk2-pytool-library:

These failures result from an unnecessarily complex change made to GetEdk2RelativePathFromAbsolutePath made in https://github.com/tianocore/edk2-pytool-library/commit/420e44a250c1b609398b839b30a2230e29016067 that tried to replace the existing logic by reusing GetContainingPackage logic. While this passed all tests at the time, additional scenarios that were once covered (but did not have associated tests) began to pop up as breaking bugs.

This PR does two things:
1. Reverts the main logical change made in https://github.com/tianocore/edk2-pytool-library/commit/420e44a250c1b609398b839b30a2230e29016067, but also provides a more elegant solution to the bug the associated PR was attempting to resolve.
2. Adds additional test coverage for the situations brought to light by this change, so that any future changes will not break the functionality again.


edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -161,9 +160,12 @@ def GetContainingPackage(self, InputPath):
                 return None
 
         # InputPath is in workspace or PackagesPath for worst case scenario.
-
-        dirpathprevious = os.path.dirname(InputPath)
-        dirpath = os.path.dirname(InputPath)
+        if os.path.isdir(InputPath):
```
spbrogan, line 163: do we have unit test coverage on both of these code paths?  I assume not given that this bug didn't get caught until integration testing.

Javagedes: I've added a test case to cover the additional codepath of if it is a folder instead of a file.

spbrogan (approved):

PR #186, opened 2022-11-03
path_utilities: Adjust GetContainingPackage() and Edk2Path behavior

### LINKED ISSUE #183: [Bug]: Unit tests can fail based on state of system temporary directory
### Contact Details

_No response_

### Describe the Bug

The `test_pp_outside_workspace()` and `test_get_containing_package_outside_workspace()` unit tests in [path_utilities_test.py](https://github.com/tianocore/edk2-pytool-library/blob/master/edk2toollib/uefi/edk2/path_utilities_test.py) pass if they do not recursively find a .dec file as they ascend up the directory hierarchy to the file system root. It does this in the user's temporary directory.

The state of the user's temporary directory is inherently unpredictable and can contain folders with any arbitrary content including .dec files.

If directories with such files exist, the unit test fails. The unit test should not be dependent on the global state of the user's temporary directory to pass/succeed. It is testing the code not the state of the user's temporary directory.

### Reproduction steps

1. Place a random directory with a .dec file in your temporary directory
2. Run the unit test test_pp_outside_workspace()

### Expected behavior

The test result should be consistent regardless of the content in the user's temporary directory.

### Execution Environment

_No response_

### Pip packages

_No response_

### Additional context

_No response_

### LINKED ISSUE #185: [Bug]: Simplify GetContainingPkg()
### Contact Details

_No response_

### Describe the Bug

`GetContainingPkg()` has confusing and undocumented behavior. This is how the function documents itself today:

```python
    def GetContainingPackage(self, InputPath: str):
        ''' Find the package this path belongs to using some Heuristic.  This isn't
        perfect but at least identifies the directory consistently.

        Note: The inputPath must be in the OS specific path form

        @param InputPath:  absolute path to a file, directory, or module. Supports both Windows and Linux like paths

        @ret Name of Package that the module is in.
        '''
```

Based on this definition, it would seem the function returns the *package* that a file path is within. However, the function returns the directory name of a file outside a package.

The function would be more useful and straightforward to simply return the name of the containing package and `None` if the path is not contained within a package.

### Reproduction steps

Call the function on a file in a directory (that is not a package) in the workspace root. The function returns the parent directory name which is seemingly arbitrary and undocumented behavior.

### Expected behavior

- Valid path outside an edk2 package -> None
- Invalid path -> well-defined exception
- Valid path in an edk2 package -> Package name

### Execution Environment

_No response_

### Pip packages

_No response_

### Additional context

_No response_

makubacki:
Fixes #183, fixes #185

Updates the Edk2Path constructor with the following changes:

- Avoid changing global object state during intermediary validation operations. If an exception were returned or execution is otherwise interrupted the `PackagePathList` and `WorkspacePath` will have predictable values only set once validation succeeds.
- Raise a `NotADirectoryError` exception on invalid directory input. This allows the caller to implement more focused exception handling.
- Update the definition of a package to be a directory with a .dec file and use that as the basis for detecting nested packages.
- Update the nested package path algorithm to fix #183 and prevent exceeding ascension in the directory hierarchy beyond the maximum root directory being compared.
- Minor documentation cleanup.

Update GetContainingPackage() with the following changes:

- Only return a containing package if a file is in a directory with a .dec file. Avoid returning directory parents outside edk2 packages. Closes #185.
- Minor documentation cleanup.

Updates both functions to remove superfluous logging and leave it to the caller to manage logging messages based on the context of the actions they are trying to perform.

Updates unit tests to align with updated function expectations.


edk2toollib/uefi/edk2/path_utilities_test.py
```diff
@@ -607,9 +607,10 @@ def test_get_relative_path_when_packages_path_list_contains_substrings(self):
         self.assertEqual(pathobj.GetEdk2RelativePathFromAbsolutePath(p), f"{ws_p_name}/module2/X64/TestFile2.c")
 
     def test_get_relative_path_when_package_path_inside_package(self):
```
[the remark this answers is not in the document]
Javagedes: Fair enough, when I made the original (broken) code, I didn't think to glob downwards to see if a package was actually in the package path. I just made the assumption that if someone was declaring a directory a package_path, it will have a package in it.

In this case, I would like to keep the test as is, but build an additional test that adds a package to the package path. so we can get some test coverage on the assert. 

Would you be willing to add that? If not, I can do it in another PR.

edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -19,64 +19,88 @@
...
+
+        for a in candidate_package_path_list:
+            if not a.is_dir():
+                if error_on_invalid_pp:
+                    raise NotADirectoryError(
+                        errno.ENOENT,
+                        os.strerror(errno.ENOENT),
+                        a.resolve())
+                else:
+                    return
```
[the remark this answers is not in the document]
Javagedes: I agree we don't want to keep the invalid package path, but there are still other valid package paths in the packagepathlist that should be kept. Your code removes all of them. I think you're confusing the user even more by removing all package paths rather than only the bad one(s). 

To me it seems like the option of not raising an error on an invalid package path has the use case of allowing a user to add package paths that may sometimes be valid and may sometimes not be valid without having to do the actual filtering (the filtering is done by Ed2Path).

What you're insinuating here is that we don't want to do that for them, and if that's the case (which I'm not against), then I don't see a legitimate reason to keep `error_on_invalid_pp` at all.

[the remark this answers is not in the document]
Javagedes: #113 Is why we have error_on_invalid_pp. Does your change to Pr_Eval resolve this so we can remove error_on_invalid_pp?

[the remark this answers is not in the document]
makubacki: The goal of the change was:

1. Inform the user in an obvious and documented way their package path is invalid - the exception
2. When the user does something wrong be consistent - always prevent them from continuing by setting zero package paths

This is simply a different philosophical approach to handling invalid user input. I'm more in favor of well-defined input expectations and preventing special handling (and complicated interfaces) from creeping in for exceptional cases. This often leads to technical debt across layers of functionality as it kind of is here now. I agree that with this approach `error_on_invalid_pp` should go away.

Given #113 explicitly allowed this behavior, I will update the code to continue to support it. Modifying this is not a major goal of this change.

edk2toollib/uefi/edk2/path_utilities_test.py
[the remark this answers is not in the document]
makubacki: I just didn't realize it wasn't covered yet; it should definitely be covered as part of these changes.

I missed the instantiation of Edk2Path in the assert in that test case and I will add it back.

edk2toollib/uefi/edk2/path_utilities.py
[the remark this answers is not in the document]
makubacki: Behavior brought back in PR update https://github.com/tianocore/edk2-pytool-library/pull/186/commits/31b16aa53b23925c6ef929798e38fe31b0834cd6

makubacki:
Note: PR gates are failing due to pydocstyle in edk2-pytool-extensions, unrelated to this change.

edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -19,64 +19,91 @@
...
+    """
+
+    def __init__(self, ws: os.PathLike, package_path_list: Iterable[os.PathLike], error_on_invalid_pp: bool = True):
+        """Constructor.
 
         Args:
-            ws: absolute path or cwd relative path of the workspace.
-            packagespathlist: list of packages path.
+            ws (os.PathLike): absolute path or cwd relative path of the workspace.
+            packagespathlist (Iterable[os.PathLike]): list of packages path.
```
spbrogan, line 32: this arg (packagespathlist) is wrong since the arg name was changed to use snake case

edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -19,64 +19,91 @@
...
+        candidate_package_path_list = []
+        for a in package_path_list:
+            if os.path.isabs(a):
+                candidate_package_path_list.append(Path(a))
             else:
-                # see if workspace relative
-                wsr = os.path.join(ws, a)
-                if (os.path.isdir(wsr)):
-                    self.PackagePathList.append(wsr)
+                wsr = Path(ws, a)
```
spbrogan, line 67: shouldn't this use self.WorkspacePath instead of ws given that ws could be cwd relative.

edk2toollib/uefi/edk2/path_utilities.py
Javagedes: @makubacki no need to fix this. I'm currently working on a pydocstyle for pytool-library and will fix it. You'll have conflicts to work through due to my changes to all documentation anyway, so you'll end up doing fixing it twice.

Javagedes: Also, pydocstyle catches issues like this, which is nice :)

edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -19,64 +19,91 @@
...
+            for i, package in enumerate(packages):
+                for j in range(i + 1, len(packages)):
+                    comp_package = packages[j]
+
+                    if (package.is_relative_to(comp_package)
+                            or comp_package.is_relative_to(package)):
+                        raise Exception(
+                            f"Nested packages not allowed. The packages "
+                            f"[{str(package)}] and [{str(comp_package)}] are "
+                            f"nested")
```
spbrogan, line 106: would appreciate a sentence or two more about this algorithm for checking for nesting.  What the intermediate  lists/dictionaries are tracking and used for.  Requires too much reverse engineering to understand.

edk2toollib/uefi/edk2/path_utilities.py
makubacki: It's a very simple change, no need for it to be wrong here. 

Updated in PR update https://github.com/tianocore/edk2-pytool-library/pull/186/commits/e21c87394eb5f15e4679f64235f30f15ac1280ca

edk2toollib/uefi/edk2/path_utilities.py
makubacki: Yeah, that's better. Updated in PR commit https://github.com/tianocore/edk2-pytool-library/pull/186/commits/e910271effdf999b4c770746f8ebd52cbd3fb0fb

edk2toollib/uefi/edk2/path_utilities.py
makubacki: Added in PR Update commit https://github.com/tianocore/edk2-pytool-library/pull/186/commits/be6ddc562fdc981aaefb31ef83e4b1e47510ab38

makubacki:
@Javagedes and @spbrogan, are you waiting on something else to approve this one?

Javagedes (approved):

Javagedes:
@makubacki Approved but there are comment conflicts since I merged the pydocstyle PR. Once the PR is passing the PR gates I'll merge.

Javagedes:
I'm also going to do a quick run of the integration checks just to make sure the change to GetContainingPackage() to return None instead of ws didn't break anything.

Javagedes:
@makubacki Integration tests passed; I'm ready to merge once the conflicts have been handled.

makubacki:
> @makubacki Integration tests passed; I'm ready to merge once the conflicts have been handled.

Conflict update pushed

PR #189, opened 2022-11-07
Enhance Edk2Path.GetContainingModules() [Rebase & FF]

### LINKED ISSUE #188: [Bug]: Edk2Path.GetContainingModules() can search beyond max limit
### Contact Details

_No response_

### Describe the Bug

The function goes up two parent levels without checking if it has passed the workspace root and all package roots. This means it will scan a directory outside the allowed search areas for an .inf file that would represent the containing module.

To exemplify the problem, ` test_get_containing_module()` will fail if a .inf file is in the user's temporary directory.

https://github.com/tianocore/edk2-pytool-library/blob/master/edk2toollib/uefi/edk2/path_utilities_test.py#L388

### Reproduction steps

Run `test_get_containing_module()` with a .inf file in your temporary directory.

### Expected behavior

`Edk2Path.GetContainingModules()` does not include invalid files in its search.

### Execution Environment

_No response_

### Pip packages

_No response_

### Additional context

_No response_

makubacki:
This series makes three main changes:

1. [Add test case to show bug in GetContainingModules()](https://github.com/tianocore/edk2-pytool-library/commit/6badaf4581bce1f53445e5899bcb481be0437e66)
   Adds `test_get_containing_module_with_infs_in_other_temp_dirs()` to
   path_utilities_test.py to show that `GetContainingModules()` will
   extend its file search for a .inf file beyond the maximum parent of
   its `Edk2Path` object instance.

   This failure would already reproduce if the user happen to have a
   .inf file in their temporary directory. This test case adds such a
   file to always show this file is returned.

   Further illustrates https://github.com/tianocore/edk2-pytool-library/issues/188

   `GetContainingModules()` will be updated to prevent this behavior in
   a future change.

2. [Verify paths in GetContainingModules() unit tests regardless of format](https://github.com/tianocore/edk2-pytool-library/commit/d6bf93c3f1a9694a67021c6d0794817f0b92bd3b)

   `test_get_containing_module()` performs most of the testing for
   the `Edk2Path.GetContainingModules()` function.

   File paths are checked based on their exact string values rather
   than whether the file path is actually the same.

   This change checks that file paths are expected while accounting
   for differences in path formats.

3. [Enhance GetContainingModules()](https://github.com/tianocore/edk2-pytool-library/commit/d07e7b9e77210a4f793b147ed3d2aab9e86666d4)

   Makes several fixes and improvements to GetContainingModules().

   Fixes #188

   These changes preserve the main behavior of the function while more
   deterministically enforcing requirements and handling additional
   scenarios.

   1. Fixes an issue where the function would look for .inf files in
      directories beyond the maximum allowed directory for the Edk2Path
      object instance.
   
      `test_get_containing_module_with_infs_in_other_temp_dirs()` now
      passes.
   
   2. Explicitly and consistently enforces that absolute paths are
      passed to the function. An exception is raised if a path is not
      absolute. In the past behavior was undefined and not considered
      in the function implementation. A test case is added to confirm
      this behavior.
   
   3. Handles n-levels of nested directories in a module directory as
      opposed to a hardcoded number of 2 (file parent and that parent)
      previously.
   
   4. Verifies the path is within an allowed path for the Edk2Path
      object (the workspace or a valid package path).
   
   5. Returns paths in a consistent and fully resolved format.
   
   6. Improves function documentation.

### DIFF
--- a/edk2toollib/uefi/edk2/path_utilities.py
+++ b/edk2toollib/uefi/edk2/path_utilities.py
@@ -245,53 +245,99 @@ class Edk2Path(object):
 
         return None
 
-    def GetContainingModules(self, InputPath: str) -> list:
-        """Find the list of modules (infs) that file path is in.
+    def GetContainingModules(self, input_path: str) -> list[str]:
+        """Find the list of modules (inf files) for a file path.
 
-        for now just assume any inf in the same dir or if none
-        then check parent dir.  If InputPath is not in the filesystem
-        this function will try to return the likely containing module
-        but if the entire module has been deleted this isn't possible.
+        Note: This function only accepts absolute paths. An exception will
+              be raised if a non-absolute path is given.
+
+        Note: If input_path does not exist in the filesystem, this function
+              will try to return the likely containing module(s) but if the
+              entire module has been deleted, this isn't possible.
+
+        - If a .inf file is given, that file is returned.
+        - Otherwise, the nearest set of .inf files (in the closest parent)
+          will be returned in a list of file path strings.
 
         Args:
-            InputPath (str): file path in the Os spefic path form
+            input_path (str): Absolute path to a file, directory, or module.
+                              Supports both Windows and Linux like paths.
 
         Returns:
-            (list): list of module inf paths in absolute form.
+            (list[str]): Absolute paths of .inf files that could be the
+                         containing module.
         """
-        self.logger.debug("GetContainingModules: %s" % InputPath)
+        input_path = Path(os.path.normcase(input_path))
+        if not input_path.is_absolute():
+            # Todo: Return a more specific exception type when
+            # https://github.com/tianocore/edk2-pytool-library/issues/184 is
+            # implemented.
+            raise Exception("Module path must be absolute.")
 
-        # if INF return self
-        if fnmatch.fnmatch(InputPath.lower(), '*.inf'):
-            return [InputPath]
+        package_paths = [Path(os.path.normcase(x)) for x in self.PackagePathList]
+        workspace_path = Path(os.path.normcase(self.WorkspacePath))
+        all_root_paths = package_paths + [workspace_path]
 
-        # Before checking the local filesystem for an INF
-        # make sure filesystem has file or at least folder
-        if not os.path.isfile(InputPath):
-            logging.debug("InputPath doesn't exist in filesystem")
+        # For each root path, find the maximum allowed root in its hierarchy.
+        maximum_root_paths = all_root_paths
+        for root_path in maximum_root_paths:
+            for other_root_path in maximum_root_paths[:]:
+                if root_path == other_root_path:
+                    continue
+                if root_path.is_relative_to(other_root_path):
+                    if len(root_path.parts) > len(other_root_path.parts):
+                        maximum_root_paths.remove(root_path)
+                    else:
+                        maximum_root_paths.remove(other_root_path)
+
+        # Verify the file path is within a valid workspace or package path
+        # directory.
+        for path in maximum_root_paths:
+            if input_path.is_relative_to(path):
+                break
+        else:
+            return []
 
         modules = []
-        # Check current dir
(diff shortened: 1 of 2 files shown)


makubacki:
@Javagedes and @spbrogan still looking for final review feedback here.

edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -230,53 +230,100 @@ def GetContainingPackage(self, InputPath):
...
+            # Given a root path is either the workspace or a package path,
+            # neither of which are a module directory, once that point is
+            # reached, all possible module candidates are exhausted.
+            current_dir = input_path.parent
+            while current_dir not in maximum_root_paths:
+                if current_dir.is_dir():
+                    current_dir_inf_files = \
+                        [f.resolve() for f in current_dir.iterdir() if
+                         f.is_file() and f.suffix.lower() == '.inf']
+                    if current_dir_inf_files:
```
Javagedes, line 319: If I'm not mistaken, It looks like current_dir_inf_files are path objects and not strings, so you are returning the wrong type.

edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -230,53 +230,100 @@ def GetContainingPackage(self, InputPath):
...
-        if (len(modules) == 0):
-            dirpath = os.path.dirname(dirpath)
-            if os.path.isdir(dirpath):
-                for f in os.listdir(dirpath):
-                    if fnmatch.fnmatch(f.lower(), '*.inf'):
-                        self.logger.debug("Found INF file in %s.  INf is: %s", dirpath, f)
-                        modules.append(os.path.join(dirpath, f))
+        if input_path.suffix == '.inf':
+            # Return the file path given since it is a module .inf file
+            modules = [input_path.resolve()]
```
Javagedes, line 290: resolve() returns a path object, rather then a string, so at the end of the function, we return a path object. 

https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve

edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -230,53 +230,100 @@ def GetContainingPackage(self, InputPath):
...
-        self.logger.debug("GetContainingModules: %s" % InputPath)
+        Args:
+            input_path (str): Absolute path to a file, directory, or module.
+                              Supports both Windows and Linux like paths.
 
-        # if INF return self
-        if fnmatch.fnmatch(InputPath.lower(), '*.inf'):
-            return [InputPath]
+        Returns:
+            (List[str]): Fully resolved absolute paths of .inf files that
```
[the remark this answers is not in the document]
makubacki: Changed to `list`

edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -230,53 +230,100 @@ def GetContainingPackage(self, InputPath):
         self.logger.info("Workspace path is : %s" % self.WorkspacePath)
         return None
 
-    def GetContainingModules(self, InputPath: str) -> list:
-        '''Find the list of modules (infs) that file path is in
-        for now just assume any inf in the same dir or if none
-        then check parent dir.  If InputPath is not in the filesystem
-        this function will try to return the likely containing module
-        but if the entire module has been deleted this isn't possible.
+    def GetContainingModules(self, input_path: str) -> list:
```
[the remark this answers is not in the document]
makubacki: Thanks, missed that it wasn't specified there

edk2toollib/uefi/edk2/path_utilities.py
makubacki: Thanks. I didn't look at the return type closely. Updated to return strings.

edk2toollib/uefi/edk2/path_utilities.py
makubacki: Same as the other instance.

Javagedes (approved):

PR #191, opened 2022-11-08
Add pydocstyle

Javagedes:
Adds PyDocStyle following the google convention.

Brings all documentation in this repo in line with this convention as the pipeline changes in edk2-pytool-extensions requires this.

This PR will be a REBASE&FF.


spbrogan:
@Javagedes

- [ ]  change the style check process to only style check the code and update developer docs to indicate.  This allows removing the noqa from those files. 
- [ ] Add basic comment to __init__.py files with either brief description of class hierarchy or reference python packaging requirements as defined https://docs.python.org/3/reference/import.html#regular-packages

spbrogan (approved):
Please make the few changes we discussed. 
Here is link to python docs about the need for __init__.py files.
https://docs.python.org/3/reference/import.html#regular-packages

PR #193, opened 2022-11-09
edk2toollib/uefi/edk2/path_utilities.py: WARNING instead of exception for nested package

### LINKED ISSUE #192: [Bug]: Raising exception for nested packages breaks some downstream builds.  Change to WARNING
### Contact Details

michael.d.kinney@intel.com

### Describe the Bug

The following commit made on April 20, 2022 added a check for nested packages (nested .dec files) and raises an exception if a nested package is detected:

https://github.com/tianocore/edk2-pytool-library/commit/420e44a250c1b609398b839b30a2230e29016067

The specific line in path_utilities.py that raises this exception is as follows:

https://github.com/tianocore/edk2-pytool-library/blob/c9aac3b6496f93c5465bbcd9894876d7d6327d97/edk2toollib/uefi/edk2/path_utilities.py#L78

A commit to the edk2 repo on Nov 1, 2022 updated pip-requirements.txt the included this change from 6 months ago.

https://github.com/tianocore/edk2/commit/b5dbf8267b2596f4a28cdb30d1f533b958ddd4ed

When this commit was added to the edk2, any downstream consumers of the edk2 repo that use pytool stuart commands and have 1 or more nested packages immediately break their build with an exception stack trace.

No information was given for this new breaking behavior.

It is clear that nested packages are not allowed by the EDK II Specifications.  However, I recommend that this issue be noted as a WARNING in the stuart logs at this time, and raise to an ERROR after all downstream consumers have had a chance to evaluate and fix the WARNING condition.

The recommended change for this issue is to change

                    raise Exception(f'Nested packages not allowed. Pkg path [{package_path}] nested in Package [{p}]')

TO:

                    self.logger.log(logging.WARNING,f'Nested packages not allowed. Pkg path [{package_path}] nested in Package [{p}]')

### Reproduction steps

1. sync with edk2 repo after 11/2/2022.
2. use stuart_ci_build on a repo with one or more nested packages
3. stuart_ci_build raises exception for nested packages

### Expected behavior

stuart_ci_build should add a WARNING to the stuart logs when a nested package is detected

### Execution Environment

Windows
Python 3.8.8

### Pip packages

Package                Version
---------------------- ---------
antlr4-python3-runtime 4.7.1
edk2-basetools         0.1.39
edk2-pytool-extensions 0.19.1
edk2-pytool-library    0.11.6
future                 0.18.2
pefile                 2022.5.30
pip                    20.2.3
PyYAML                 6.0
semantic-version       2.10.0
setuptools             49.2.1

### Additional context

_No response_

mdkinney:
Resolves #192 

Update the check for nested packages to generate a WARNING instead of raising an exception.

This check is valuable and all nested packages should be removed. By downgrading this check to a WARNING, the downstream consumers of edk2 can be notified of the issue and be given time to resolve without breaking stuart builds.

Signed-off-by: Michael D Kinney <michael.d.kinney@intel.com>

### DIFF
--- a/edk2toollib/uefi/edk2/path_utilities.py
+++ b/edk2toollib/uefi/edk2/path_utilities.py
@@ -75,7 +75,7 @@ class Edk2Path(object):
                 if str(p).lower() == str(ws).lower():
                     break
                 if len(glob.glob(f'{p}/*dec')) != 0:
-                    raise Exception(f'Nested packages not allowed. Pkg path [{package_path}] nested in Package [{p}]')
+                    self.logger.log(logging.WARNING,f'Nested packages not allowed. Pkg path [{package_path}] nested in Package [{p}]')
                 p = p.parent
 
     def GetEdk2RelativePathFromAbsolutePath(self, abspath):
--- a/edk2toollib/uefi/edk2/path_utilities.py
+++ b/edk2toollib/uefi/edk2/path_utilities.py
@@ -75,7 +75,8 @@ class Edk2Path(object):
                 if str(p).lower() == str(ws).lower():
                     break
                 if len(glob.glob(f'{p}/*dec')) != 0:
-                    self.logger.log(logging.WARNING,f'Nested packages not allowed. Pkg path [{package_path}] nested in Package [{p}]')
+                    self.logger.log(logging.WARNING,
+                                    f'Nested packages not allowed. Pkg path [{package_path}] nested in Package [{p}]')
                 p = p.parent
 
     def GetEdk2RelativePathFromAbsolutePath(self, abspath):
--- a/edk2toollib/uefi/edk2/path_utilities_test.py
+++ b/edk2toollib/uefi/edk2/path_utilities_test.py
@@ -13,6 +13,7 @@ import sys
 import tempfile
 import shutil
 from edk2toollib.uefi.edk2.path_utilities import Edk2Path
+import pytest
 
 
 class PathUtilitiesTest(unittest.TestCase):
@@ -606,6 +607,7 @@ class PathUtilitiesTest(unittest.TestCase):
         p = os.path.join(ws_pkg_abs, "module2", "X64", "TestFile2.c")
         self.assertEqual(pathobj.GetEdk2RelativePathFromAbsolutePath(p), f"{ws_p_name}/module2/X64/TestFile2.c")
 
+    @pytest.mark.skip(reason="disable test until all nested package issues resolved")
     def test_get_relative_path_when_package_path_inside_package(self):
         ''' test usage of GetEdk2RelativePathFromAbsolutePath when a package_path directory is
         inside the subfolders of a package. Should raise an exception.
--- a/edk2toollib/uefi/edk2/path_utilities_test.py
+++ b/edk2toollib/uefi/edk2/path_utilities_test.py
@@ -13,7 +13,6 @@ import sys
 import tempfile
 import shutil
 from edk2toollib.uefi.edk2.path_utilities import Edk2Path
-import pytest
 
 
 class PathUtilitiesTest(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


BobCF (approved):

BobCF (approved):

makubacki:
@mdkinney - I think the exception is kind of nice because it requires explicit action to acknowledge you have a problem and either defer it (by handling the exception) or fix it. That action is tracked in the source code.

Why can't the calling code just handle the exception?

mdkinney:
If init of Edk2Path object raises an exception, then the object is not created, which means even if caught in except clause, the stuart command can not continue.  Also, creation of Edk2Path object can fail for many reasons, and there is no return value from object creation to know what problem was identified.

makubacki:
> @makubacki - If init of Edk2Path object raises an exception, then the object is not created, which means even if caught in except clause, the stuart command can not continue. Also, creation of Edk2Path object can fail for many reasons, and there is no return value from object creation to know what problem was identified.
> 

I forgot this was in the object constructor so it would impact the object being created.

> Also, the Edk2Path object creation contains some checks that are fatal error conditions where execution of stuart can not continue. These are Workspace is not a directory and an item in PackagesPathList is not a directory. It makes sense for fatal conditions to raise an exception. Nested packages/dec files are not conformant with EDK II Specs, but presence if nested packages/dec files does not prevent stuart from running. So that is not a condition that should raise an exception. Logging a warning now makes more sense.

I believe the Edk2Path constructor is allowed to consider any violation of fundamental edk2 path rules that complicate its overall logic or operation (i.e. object function implementation) worthy of an exception. We log many warnings that don't get fixed, and loggers can be filtered to entirely ignore warnings.

We believe a global flag (build environment variable) that when set ignores breaking on nested package would allow us to maintain the default behavior we want - not allow nested packages and make it obvious in the code that sets the flag it is acknowledging that it is allowing nested packages.

Would that work for you?

mdkinney:
I would prefer the current default behavior be WARNING.  And once all downstreams fix nested packages make the default ERROR with an option to treat as a WARNING.

This change was added to edk2 repo with pytool version update with no notice to the EDK II community of this new rule enforcement with no transition period.  We need to plan the introduction of new rule enforcements that always provide a window with rule active without break builds, and then later enforce with breaking builds.

makubacki:
> I would prefer the current default behavior be WARNING. And once all downstreams fix nested packages make the default ERROR with an option to treat as a WARNING.
> 
> This change was added to edk2 repo with pytool version update with no notice to the EDK II community of this new rule enforcement with no transition period. We need to plan the introduction of new rule enforcements that always provide a window with rule active without break builds, and then later enforce with breaking builds.

Okay, I understand the adoption/breakage concern and we want to increase stuart adoption and overall usefulness. I'll leave it up to others to make the decision.

Regarding the downstream making fixes, how do you think we track that? I haven't looked back at the versioning details, but semantic versioning is supposed to help with changes like this (rolling major version to indicate incompatible API changes).

makubacki:
@mdkinney, @spbrogan , @BobCF 

This is what we were thinking: [Edk2Path: Add an env variable to allow nested packages](https://github.com/tianocore/edk2-pytool-library/pull/199)

Does this work for you?

mdkinney:
I still think the logic is reversed.  This is still a breaking change for devs and CI agents that now have to be updated with no notification to set an env var.

If you changed it to an env var to enable nested package checks and with a plan to not require the variable in the future, that would be better.

Also, why is this nested package check in stuart and not a CI plugin?  Does the presence of a nested package actually prevent stuart commands from running correctly?  If not, then perhaps this check should be removed at this scope and moved into a pluigin.

makubacki:
@mdkinney, @spbrogan, and I met offline and found that https://github.com/tianocore/edk2-pytool-library/pull/199 will work for Mike's use case. I believe we can close this PR now that https://github.com/tianocore/edk2-pytool-library/pull/199 is merged.

PR #199, opened 2022-11-09
Edk2Path: Add an env variable to allow nested packages

makubacki:
Adds an environment variable called `PYTOOL_TEMPORARILY_IGNORE_NESTED_EDK_PACKAGES`
that a platform can set to `"true"` to allow nested packages.

Nested packages are a violation of EDK II package rules and should
be fixed. This is provided for backward compatibility and temporary
use until the variable is no longer needed.

### DIFF
--- a/edk2toollib/uefi/edk2/path_utilities.py
+++ b/edk2toollib/uefi/edk2/path_utilities.py
@@ -100,9 +100,13 @@ class Edk2Path(object):
             package_path_packages[package_path] = \
                 [Path(p).parent for p in package_path.glob('**/*.dec')]
 
+        # Note: The ability to ignore this function raising an exception on
+        #       nested packages is temporary. Do not plan on this variable
+        #       being available long-term and try to resolve the nested
+        #       packages problem right away.
         ignore_nested_packages = False
-        if "STUART_IGNORE_EDK_NESTED_PACKAGES" in os.environ and \
-            os.environ["STUART_IGNORE_EDK_NESTED_PACKAGES"].strip().lower() == \
+        if "PYTOOL_TEMPORARILY_IGNORE_NESTED_EDK_PACKAGES" in os.environ and \
+            os.environ["PYTOOL_TEMPORARILY_IGNORE_NESTED_EDK_PACKAGES"].strip().lower() == \
                 "true":
             ignore_nested_packages = True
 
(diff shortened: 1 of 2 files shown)


edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -100,17 +100,31 @@ def __init__(self, ws: os.PathLike, package_path_list: Iterable[os.PathLike],
             package_path_packages[package_path] = \
                 [Path(p).parent for p in package_path.glob('**/*.dec')]
 
+        ignore_nested_packages = False
```
spbrogan, line 103: should we put comments around this code block that reference this will be removed.

edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -100,17 +100,31 @@ def __init__(self, ws: os.PathLike, package_path_list: Iterable[os.PathLike],
...
+        if "STUART_IGNORE_EDK_NESTED_PACKAGES" in os.environ and \
+            os.environ["STUART_IGNORE_EDK_NESTED_PACKAGES"].strip().lower() == \
```
spbrogan, line 105: should the variable name mention something about temporary.  Given the pytool library can be used with more than stuart i think it makes sense to call this pytool instead of stuart. How about something like `PYTOOL_TEMPORARLY_IGNORE_NESTED_EDK_PACKAGES`

edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -100,17 +100,31 @@ def __init__(self, ws: os.PathLike, package_path_list: Iterable[os.PathLike],
...
-                        raise Exception(
-                            f"Nested packages not allowed. The packages "
-                            f"[{str(package)}] and [{str(comp_package)}] are "
-                            f"nested")
+                        if ignore_nested_packages:
+                            self.logger.log(
+                                logging.WARNING,
+                                f"Nested packages not allowed. The packages "
+                                f"[{str(package)}] and [{str(comp_package)}] are "
+                                f"nested.")
```
spbrogan, line 120: Should this printed text also indicate that user has environment variable set to ignore this error.  

Should it also mention that some pytools features may not work as expected due to ignoring this requirement?

edk2toollib/uefi/edk2/test_path_utilities.py
```diff
@@ -811,9 +811,15 @@ def test_get_relative_path_with_nested_packages(self):
         pp2_name = "PPTestPkg2"
         self._make_edk2_package_helper(folder_pp2_abs, pp2_name)
 
+        # Nested packages should raise an exception by default
         self.assertRaises(Exception, Edk2Path, folder_ws_abs, [folder_pp1_abs])
         self.assertRaises(Exception, Edk2Path, folder_ws_abs, [folder_pp1_abs, folder_pp2_abs])
 
+        # Nested packages should no longer raise an exception
```
spbrogan, line 818: a brief comment here to say this this can be removed once the workaround feature is removed will make it easier in future.

edk2toollib/uefi/edk2/path_utilities.py
makubacki: Added in PR commit https://github.com/tianocore/edk2-pytool-library/pull/199/commits/faf600ea5c0691f6f30c5ffdc9650750c245a90c

edk2toollib/uefi/edk2/path_utilities.py
makubacki: Added in PR commit https://github.com/tianocore/edk2-pytool-library/pull/199/commits/faf600ea5c0691f6f30c5ffdc9650750c245a90c

edk2toollib/uefi/edk2/test_path_utilities.py
makubacki: Added in PR commit https://github.com/tianocore/edk2-pytool-library/pull/199/commits/faf600ea5c0691f6f30c5ffdc9650750c245a90c

edk2toollib/uefi/edk2/path_utilities.py
makubacki: Updated variable name in PR commit https://github.com/tianocore/edk2-pytool-library/pull/199/commits/faf600ea5c0691f6f30c5ffdc9650750c245a90c

spbrogan (approved):

Issue #201, opened 2022-11-10
[Feature]: Clarify the supported versions of python and add tests for each


spbrogan:
### What does the feature solve?

In this project it is not clear what versions of python are required, supported, and tested against.  This makes it harder for consumers of this project to know their requirements are.  It also makes it harder for the developers of this project as they don't know what language features they are allowed to use.  

### Describe the solution

1. Add documentation for the supported versions of python and the rationale and process.  
2. Update documentation, Pypi metadata to be accurate.
3. Add PR gate tests for the supported versions of python
4. Add CI gates and status badges for each supported version

### Additional context

As pytools  usage expands it is critical to communicate these things as well as provide roadmap to indicate how these change as python releases new versions.

Javagedes:
@makubacki @spbrogan I vote to only garuntee support for last three major versions (so 3.9,3.10, 3.11). When 3.12 comes out we bump it to 3.10, 3.11, 3.12. I can setup a matrix for pipelines that changes the version and runs all tests.

makubacki:
> @makubacki @spbrogan I vote to only garuntee support for last three major versions (so 3.9,3.10, 3.11). When 3.12 comes out we bump it to 3.10, 3.11, 3.12. I can setup a matrix for pipelines that changes the version and runs all tests.

I agree with n-3. This seems to capture the minimum version I've seen required in some environments right now (3.9). We've also discussed n-2 but I think n-3 is a good starting point.

Javagedes:
Adding required python version for pypi in #223, but this does not solve the entire Issue.

Javagedes:
Pushing this milestone back. This will be marked complete when I upgrade from pipelines to github actions similar to what was done in edk2-pytool-extensions. Holding off on this until we have something similar to mu_devops.

PR #202, opened 2022-11-10
bug_report.yml: Add supported Python versions

makubacki:
Lists the Python versions supported for bug fixes in the bug reporting form.

This is to make it obvious what Python versions are officially supported for bug fixes when a bug is filed.

Contributes to https://github.com/tianocore/edk2-pytool-library/issues/201

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -41,11 +41,23 @@ body:
       description: "A clear and concise description of what you expected to happen."
     validations:
       required: true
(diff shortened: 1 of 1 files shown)


Javagedes:
I would like to add the pipeline change in before we update the supported version as our pipeline currently runs 3.10.6.

makubacki:
> I would like to add the pipeline change in before we update the supported version as our pipeline currently runs 3.10.6.

Sounds good. Just wanted to get this piece in place. It can go in after that.

spbrogan (approved):

PR #206, opened 2022-11-11
Add GetEdk2RelativePathFromAbsolutePath() POSIX test case

makubacki:
Adds a test case to verify the function returns POSIX paths as expected.


makubacki:
This was a test I wrote while checking various function return values. Decided we might as well check it in.

Javagedes:
Does this close #207 ?

makubacki:
> Does this close #207 ?

No. I have something that can close it soon, but it's not high priority since the function follows its documented interface.

Javagedes (approved):

PR #214, opened 2022-12-07
Add automatic Documentation Generation [REBASE&FF]

Javagedes:
Three separate commits to:

1. Set up automatic documentation generation
2. Move tests to a test folder
3. Fix incorrect style for generating tips, hints, warnings, alerts
4. Add a documentation Issue template

NOTE: This does not contain the logic to push the static html to readthedocs to be hosted. That PR will happen once I get both edk2toolext and edk2toollib finished so I can add the final pipeline check without blocking edk2toollib.

download


makubacki:
This change helps highlight how disorganized the sections are. We definitely need to clean them up in the future. But this is a good baseline to do that work on.

makubacki:
This change helps highlight how disorganized the sections are. We definitely need to clean them up in the future. But this is a good baseline to do that work on.

PR #221, opened 2023-01-04
Improve error message in locate_tools when FindWithVsWhere fails

Javagedes:
A poor error message when vswhere fails to find the vs path does not provide adequate information to allow the user to attempt to solve the problem. This provides additional logging.error information to the user that they might need to verify the specified vs version exists. If no vs version was specified in the call, then information is provided to the user that they may need to specify the vs version (as there could be a conflict).

Breaking change as additional exception types were introduced.


edk2toollib/windows/locate_tools.py
```diff
@@ -174,7 +174,11 @@ def QueryVcVariables(keys: list, arch: str = None, product: str = None, vs_versi
     result = {}
     ret, vs_path = FindWithVsWhere(product, vs_version)
     if ret != 0 or vs_path is None:
-        logging.warning("We didn't find VS path or otherwise failed to invoke vsWhere")
+        logging.error("We didn't find VS path or otherwise failed to invoke vsWhere")
```
spbrogan, line 177: if this is trying to compare against none you should check if "is None".  as the below conditional will also trigger if vs_version == 0

spbrogan (approved):

edk2toollib/windows/locate_tools.py
Javagedes: now checks via `if vs_version is not None:` as suggested

PR #227, opened 2023-01-12
Add support for ignoring known-bad packages

joschock:
This adds an environment variable to allow the nested package check to be demoted to an INFO-level log for packages that are known to be bad and cannot be immediately remediated. This avoids a large number of warning messages that would otherwise make it more likely for developers to ignore warnings entirely.

### DIFF
--- a/edk2toollib/tests/test_path_utilities.py
+++ b/edk2toollib/tests/test_path_utilities.py
@@ -947,6 +947,23 @@ class PathUtilitiesTest(unittest.TestCase):
         # Remove the environment variable now that the test above is complete
         os.environ.pop("PYTOOL_TEMPORARILY_IGNORE_NESTED_EDK_PACKAGES")
 
+        # Nested packages should no longer raise an exception if explicitly
+        # marked as known-bad.
+        os.environ["PYTOOL_IGNORE_KNOWN_BAD_NESTED_PACKAGES"] = "SomeOtherPkg,PPTestPkg1"
+        Edk2Path(folder_ws_abs, [folder_pp1_abs])
+        Edk2Path(folder_ws_abs, [folder_pp1_abs, folder_pp2_abs])
+
+        os.environ["PYTOOL_IGNORE_KNOWN_BAD_NESTED_PACKAGES"] = "SomeOtherPkg,PPTestPkg2"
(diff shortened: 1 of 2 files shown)


makubacki (approved):

joschock:
> 

Well, we need those packages for our project; so removing them from Packages Path would break our build.

Javagedes:
@joschock I misunderstood what this was doing. I have to say, I'm not a huge fan because we purposefully added these warnings so that people will go in and fix the nested path rather than ignore it. 

Being able to make nested package paths only a warning is a temporary change to allow platforms to fix their nested packages before we turn it into a permanent failure. Adding the ability to further suppress these warnings just allows platforms to ignore the issue.

There is an argument that you are acknowledging the nested platform by setting the environment variable manually... until one just makes it automatically set in the platform build file, or elsewhere. 

I'll let @spbrogan be the final judge though.

joschock:
> @joschock I misunderstood what this was doing. I have to say, I'm not a huge fan because we purposefully added these warnings so that people will go in and fix the nested path rather than ignore it.
> 
> Being able to make nested package paths only a warning is a temporary change to allow platforms to fix their nested packages before we turn it into a permanent failure. Adding the ability to further suppress these warnings just allows platforms to ignore the issue.
> 
> There is an argument that you are acknowledging the nested platform by setting the environment variable manually... until one just makes it automatically set in the platform build file, or elsewhere.
> 
> I'll let @spbrogan be the final judge though.

I agree in principle - nested packages are bad and should be fixed. My issue is that these packages are largely beyond our control. Our partner is unlikely to fix this in a timely fashion. If we override their code to resolve this (which would be a fairly extensive modification, given the pervasiveness of this bad practice in their code), then we have to carry and maintain those modifications and that's an ongoing engineering cost for us that we pay at every integration - it would mean that our file tree layouts would fundamentally differ, so it would break merging pretty badly. If we don't fix them and don't take this change, then we end up with several thousand warning messages to the console on every build; this trains all of our developers to ignore warning messages - meaning that warning messages basically become ignorable noise and all other errors that we could actually do something about would vanish into the maelstrom of nested package warnings.  Essentially what I'm proposing with this PR is allowing platform integrators to mark packages as known-bad, and demote the corresponding errors from WARN to INFO level.

Javagedes (approved):

PR #240, opened 2023-01-30
Only download vswhere if on windows

### LINKED ISSUE #235: [Bug]: fails to install when offline
### Contact Details

just add a comment, and let github send out notifications 😎

### Describe the Bug

Installing the package without internet connection (standard procedure for linux distro packaging) fails because the installer wants download vswhere.exe.

Silently downloading stuff is bad practice.

In this specific case it also looks rather pointless to do this on linux machines, they typically wouldn't be able to run the binary anyway.

### Reproduction steps

1. clone git repo
2. disable network
3. try install package

### Expected behavior

install works

Python 3.11

### Execution Environment

Fedora 37

### Pip packages

_No response_

### Additional context

_No response_

Javagedes:
This updates the setup.py script used for installing edk2-pytool-extensions as a pypi module to only install vswhere if we are on a windows device.

Closes #235

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,16 +12,23 @@ from setuptools.command.sdist import sdist
 from setuptools.command.install import install
 from setuptools.command.develop import develop
 from edk2toollib.windows.locate_tools import _DownloadVsWhere
+from edk2toollib.utility_functions import GetHostInfo
 
 with open("readme.md", "r") as fh:
     long_description = fh.read()
+
+
+def _download_vswhere_if_windows(): # noqa
+    """Downloads vswhere only if on the windows OS."""
+    if GetHostInfo().os == "Windows":
+        _DownloadVsWhere()
 
 
 class PostSdistCommand(sdist): # noqa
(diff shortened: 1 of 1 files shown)


apop5 (approved):

spbrogan (approved):
Please raise exception and/or log a clear message about the vswhere api being Windows only.  
https://github.com/tianocore/edk2-pytool-library/blob/master/edk2toollib/windows/locate_tools.py

Javagedes:
> Please raise exception and/or log a clear message about the vswhere api being Windows only. https://github.com/tianocore/edk2-pytool-library/blob/master/edk2toollib/windows/locate_tools.py

This is being resolved in #221. GetVsWherePath now raises an error if the os is not Windows. This will propagate through `FindWithVsWhere()`, `QueryVcVariables()`, and `FindToolInWinSdk()` as an EnvironmentError with a well-defined description: "VsWhere cannot be used on a non-windows system."

Issue #245, opened 2023-02-01
[Feature]: Update authenticated_variables_structure_support and wincert


Flickdm:
### What does the feature solve?

The following changes would be useful for building and describing a UEFI Authenticated variable 

1.  Wincert should be able to parse the asn.1 signature instead of only the win_certificate since wincert is the header for the signature data.
2. Wincert should be able to encode its own structure.
3. EFITime should be able to encode its own structure.
4. There should be a AuthenticatedVariableBuilder that builds an authenticated variable class.
6. The output of any of the print functions should be able to be directed to a file

### Describe the solution

Implement class EfiVariableAuthentication2Builder(object):
Update Wincert
Update EfiTime
Update EfiVariableAuthentication2

Support python3 string syntax

Flickdm:
See https://github.com/tianocore/edk2-pytool-library/pull/246

Flickdm:
unit test and code depends on #244

PR #246, opened 2023-02-01
Updated authenticated_variables_structure

Flickdm:


Flickdm:
Closes #245

edk2toollib/uefi/authenticated_variables_structure_support.py
```diff
@@ -53,32 +69,33 @@ def __init__(self,
             (Exception): Invalid Parameters
         """
         if (decodefs is not None):
-            self.PopulateFromFileStream(decodefs, decodesize)
+            self.decode(decodefs, decodesize)
         elif (createfs is not None):
             # create a new one
-            self.SignatureOwner = sigowner
-            start = createfs.tell()  # should be 0 but maybe this filestream has other things at the head
+            self.signature_owner = sigowner
```
Javagedes, line 75: I think with these changes you'll need to do python's getter/setter changes so you can do deprecation warnings on the variables that you are changing the name of. Hopefully no one is accessing this data directly but you just never know.

https://www.geeksforgeeks.org/getter-and-setter-in-python/

edk2toollib/uefi/wincert.py
```diff
@@ -52,41 +57,64 @@ class WinCertPkcs1(object):
 
     def __init__(self, filestream=None):
         """Inits the object."""
-        if (filestream is None):
+        if filestream is None:
             self.Hdr_dwLength = WinCertPkcs1.STATIC_STRUCT_SIZE
             self.Hdr_wRevision = WinCert.REVISION
             self.Hdr_wCertificateType = WinCert.WIN_CERT_TYPE_EFI_PKCS115
-            self.HashAlgorithm = None
```
Javagedes, line 59: same as the previous. We either need to using getter / setter to throw warnings for these values, or I can do a minor level release and tell everyone we have switched from CamelCase to SnakeCase for these class attributes and that they should do the methods too, but that for now those are only deprecated

Javagedes (approved):
This all looks good. I see a fair amount of duplication in methods. I noticed there are a lot of methods named the same between classes; I was wondering if you thought if there was enough duplication here to create a base class / subclass architecture to reduce some of it? It could also just be that there's a lot of functions named the same, but work differently internally.

edk2toollib/uefi/authenticated_variables_structure_support.py
```diff
@@ -738,117 +999,223 @@ def GetCanonicalAndDupes(self):
...
         return (canonicalDb, duplicatesDb)
 
-    def GetCanonical(self):
+    def GetCanonicalAndDupes(self):
+        """Compute and return a tuple containing both a canonicalized database & a database of duplicates.
+
+        Returns:
+            (Tuple[EfiSignatureDatabase, EfiSignatureDatabase]): (canonical, duplicates)
+
+        NOTE:
```
Javagedes: for notes, we use:

```
!!! note
    notes here
```

pyproject.toml
```diff
@@ -9,7 +9,11 @@ dynamic = ["version"]
 description = "Python library supporting UEFI EDK2 firmware development"
 readme = {file = "readme.md", content-type = "text/markdown"}
 license = {file = "LICENSE"}
-dependencies = []
+dependencies = [
+    "pyasn1==0.4.8",
+    "pyasn1-modules==0.2.8",
+    "cryptography==39.0.0",
```
Javagedes, line 15: As these true dependencies (not dev dependencies) we want to be as flexible as possible with requirements. Typically we do this via `>=` rather than `==`.

Also please require 39.0.1 or greater for cryptography: see https://cryptography.io/en/latest/changelog/#v39-0-1

edk2toollib/uefi/authenticated_variables_structure_support.py
Javagedes: This shows up as pretty little alert box in the docs

PR #269, opened 2023-02-27
Update vscode settings

Javagedes:
Transitions vscode default test settings to pytest rather then unittests to match what is done for CI via github actions.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,13 +1,9 @@
 {
-  "python.testing.unittestArgs": [
-    "-v",
-    "-s",
+  "python.testing.pytestEnabled": true,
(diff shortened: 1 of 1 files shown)


makubacki:
My understanding is that pytest should retain support for all the unittest tests in Test Explorer. On a quick look of pulling down the branch that looks to be the case.

makubacki (approved):

Javagedes:
> My understanding is that pytest should retain support for all the unittest tests in Test Explorer. On a quick look of pulling down the branch that looks to be the case.

Yes no tests using unittest are affected and need to be changed. This just adds an additional (and in my opinion, better) testing architecture due to the use of fixtures.

PR #286, opened 2023-03-31
base_parser.py: Allow Edk2PathObj to be passed

Javagedes:
base_parser.py: BaseParser was identified as a class that instantiates an Edk2Path object but could also be instantiated itself many times throughout the build process (once per INF). As a part of https://github.com/tianocore/edk2-pytool-extensions/issues/446, we are working to minimize the amount of Edk2PathObj instantiations due to the slow verification of workspace paths that occur.

This change adds a new method, `SetEdk2Path()` which is a drop in replacement for `SetBaseAbsPath()` and `SetPackagePaths()`.

Edk2Path's constructure requires both the workspace path (BaseAbsPath) and the Packages path list. Due to this, we can instantiate the Edk2Path object once, and pass it to each `BaseParser()` instance, changing the amount of calls from 2*num_of_parsers to 1.

Additionally, `SetBaseAbsPath()` and `SetPackagePaths()` have been deprecated.

### DIFF
--- a/edk2toollib/uefi/edk2/parsers/base_parser.py
+++ b/edk2toollib/uefi/edk2/parsers/base_parser.py
@@ -49,10 +49,32 @@ class BaseParser(object):
     #
 
     def SetEdk2Path(self, pathobj: path_utilities.Edk2Path):
-        """Sets the attribute RootPath.
+        """Sets the internal attribute Edk2PathUtil.
 
         !!! note
-            This is a drop in replacement for SetBaseAbsPath and SetPackagePaths
+            This is a drop in replacement for SetBaseAbsPath and SetPackagePaths as it will asssign both RootPath
+            and PPs using the Edk2Path object attributes WorkspacePath and PackagePathList respectively.
+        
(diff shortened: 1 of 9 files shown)


edk2toollib/uefi/edk2/parsers/base_parser.py
```diff
@@ -47,6 +48,22 @@ def __init__(self, log="BaseParser"):
     # For include files set the base root path
     #
 
+    def SetEdk2Path(self, pathobj: path_utilities.Edk2Path):
+        """Sets the attribute RootPath.
```
spbrogan, line 52: better comments here about what is required.  
Reference to deprecated functions as the docs comment wont be much good when those are gone.

spbrogan (approved):
suggested a minor update

edk2toollib/uefi/edk2/parsers/base_parser.py
Javagedes: Good call. I've update the docs with integration instructions and will do the same when announcing the deprecation (but not yet removal) of the two functions in the upcoming release.

PR #290, opened 2023-04-04
utility_functions.py - bugfix encoding when locale information missing

Javagedes:
If locality information is missing on the system, `locale.getdefaultlocale()` fails to return a valid encoding and thus any `RunCmd()` call will fail. This adds in the logic to default to utf-8 if locality information returns None for the encoding.


apop5 (approved):

apop5:
The change worked for the build system that did not have any locale settings.

kuqin12 (approved):

cfernald (approved):

Issue #291, opened 2023-04-05
[Bug]: Current capsule generation DestDir follows deprecated convention and needs to be updated to match latest guidance.


joschock:
### Contact Details

joschock@microsoft.com

### Describe the Bug

Current capsule INF generation doesn't follow latest guidelines from https://learn.microsoft.com/en-us/windows-hardware/drivers/bringup/authoring-an-update-driver-package with respect to DestinationDir. 

See for example here, where %DIRID_WINDOWS% is still used:

https://github.com/tianocore/edk2-pytool-library/blob/c9d1581cd8b428777cdaa8c369a3cec25595408f/edk2toollib/windows/capsule/inf_generator2.py#L334

### Reproduction steps

1. Generate a capsule using these tools.
2. Open the INF

### Expected behavior

Generated INF conforms to latest guidelines heres: https://learn.microsoft.com/en-us/windows-hardware/drivers/bringup/authoring-an-update-driver-package

### What Python version are you using?

Python 3.9, Python 3.10, Python 3.11

### Execution Environment

Normal UEFI build environments.

### Pip packages

Not relevant. 

### Additional context

If desired, a flag could be used to provide backward compatibility.

PR #315, opened 2023-05-01
inf_generator: update DefaultDestDir

Javagedes:
The required DefaultDestDir was changed from `%DIRID_WINDOWS%,Firmware\{6bd4efb9-23cc-4b4a-ac37-016517413e9a}` to `13` with the release of Windows 10 1803 but was never updated in either inf_generator.py or inf_generator2.py. As Windows 10 1803 has been EOL since November 12, 2019,  I see no reason to support the previous requirements and will transition to the new value.

Thanks @joschock for bringing this to light in #291

### DIFF
--- a/edk2toollib/windows/capsule/inf_generator2.py
+++ b/edk2toollib/windows/capsule/inf_generator2.py
@@ -331,7 +331,7 @@ class InfSourceFiles(object):
         outstr += Files
         outstr += textwrap.dedent("""
             [DestinationDirs]
-            DefaultDestDir = %DIRID_WINDOWS%,Firmware ; %SystemRoot%\\Firmware
+            DefaultDestDir = 13
 
             """)
         return outstr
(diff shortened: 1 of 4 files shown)


edk2toollib/windows/capsule/inf_generator2.py
```diff
@@ -331,7 +331,7 @@ def __str__(self) -> str:
         outstr += Files
         outstr += textwrap.dedent("""
             [DestinationDirs]
-            DefaultDestDir = %DIRID_WINDOWS%,Firmware ; %SystemRoot%\\Firmware
+            DefaultDestDir = 13
```
antklein, line 334: I believe line 240 needs to be updated as well.

https://learn.microsoft.com/en-us/windows-hardware/drivers/bringup/authoring-an-update-driver-package
HKR,,FirmwareFilename,,%13%\firmware.bin
; Prior to Windows 10 1803, the above should instead be:
; HKR,,FirmwareFilename,,{6bd4efb9-23cc-4b4a-ac37-016517413e9a}\firmware.bin

Javagedes: Ah yes. Thank you; I will update now.

antklein (approved):

makubacki (approved):

PR #320, opened 2023-05-04
locate_tools: add detailed warnings for vcvarsall.bat errors

Javagedes:
Adds additional logging to `QueryVcVariables()` when calling vcvarsall.bat. command prompt arguments are limited to 8191 characters, which includes ENV variables. On top of that, any ENV variable that exceeds 8191 characters itself is not passed to a command.

This causes two different errors when confusing:

1. When the total env is greater than 8191 characters (but a specific variable is not), the command fails with an ambigious error `Exception: The input line is too long.
The syntax of the command is incorrect.`

2. When a specific variable is greater than 8191 characters (commonly the PATH variable), it is not passed to the command. This means that the command does not fail, however specifically for vcvarsall.bat, we might fail to find what we are searching for because the folder it resides in could set in PATH.

This commit adds the following:

1. When the PATH variable is greater than 8191 characters, a warning will notify the user that PATH will not be included when searching for their keys.
2. When an exception occurs stating that the input line is too long, we expand on the exception to provide a clear message as to why the error is occurring.


edk2toollib/windows/locate_tools.py
```diff
@@ -203,13 +203,25 @@ def QueryVcVariables(keys: list, arch: str = None, product: str = None, vs_versi
         logging.error(err_msg)
         raise ValueError(err_msg)
 
+    if len(os.environ['PATH']) > 8191:
```
makubacki, line 206: Are you planning to address the code coverage warnings?

Javagedes: Sure, I can add some tests. This is not a difficult scenario to reproduce.

Javagedes: @makubacki Tests have been added to cover the added error handling.

Flickdm (approved):

tests.unit/test_locate_tools.py
```diff
@@ -99,3 +101,44 @@ def test_QueryVcVariablesWithVariableNotFound(self):
...
+    
+    When calling a .bat file, the environment is passed in, however any environment variable greater
+    than 8191 is quietly ignored. This can sometimes happen with the PATH variable, but we almost
+    always need the PATH env variable when querying VcVariables, so we want to warn when the user
+    when this happens.
+    """
+    with caplog.at_level(logging.WARNING):
+        keys = ["WindowsSDKVersion"]
+        locate_tools.QueryVcVariables(keys)
+        assert (len(caplog.records)) == 0
```
makubacki, line 117: Was there a situation that led to using an assert instead of one of the `TestCase` class assert methods here?

makubacki (approved):

tests.unit/test_locate_tools.py
Javagedes: This test is using Pytest instead of unittest. Pytest provides more functionality then in it test, which I utilize for this test (the caplog).

As I am utilizing pytest, this function cannot (and is not) a part of the above class, so I'm unable to utilize the assert methods it provides as it requires self.

In a perfect world, all tests would be switched from unittest to pytest, but that is a lot of effort when the unit tests run correctly and are good, valid tests.

makubacki: That's fine I was just looking at the test code and missed the framework switch.

PR #372, opened 2023-07-20
path_utilities: Use pathlib

### LINKED ISSUE #207: [Bug]: GetEdk2RelativePathFromAbsolutePath() Can Silently Return Non-POSIX Path
### Contact Details

_No response_

### Describe the Bug

GetEdk2RelativePathFromAbsolutePath() is currently defined as follows:

```python
        """Given an absolute path return a edk2 path relative to workspace or packagespath.

        Note: absolute path must be in the OS specific path form
        Note: the relative path will be in POSIX-like path form

        Args:
            abspath (os.PathLike): absolute path to a file or directory. Path must contain OS specific separator.

        Returns:
            (os.PathLike): POSIX-like relative path to workspace or packagespath
            (None): abspath is none
            (None): path is not valid
        """
```

It requires an OS-specific path (to the current OS) to be given as input. However, it does not return an exception or other fail in a well-defined way if path given does not conform to the current OS. It likely returns a non-POSIX path.

So, this is not a bug with the function in the case the user passes the expected path format but a bug in validating the function input values.

### Reproduction steps

The following code demonstrates the problem if run on Linux since it passes a Windows style path to the function.

```
        p = f"{ws_pkg_abs}\\module_2\\X64\\TestFile.inf"
        expected_rel_from_abs_path = PurePath(os.path.join(ws_p_name, "module_2", "X64", "TestFile.inf")).as_posix()
        actual_rel_from_abs_path = pathobj.GetEdk2RelativePathFromAbsolutePath(p)
>       self.assertEqual(expected_rel_from_abs_path, actual_rel_from_abs_path)
E       AssertionError: 'WSTestPkg/module_2/X64/TestFile.inf' != 'WSTestPkg\\module_2\\X64\\TestFile.inf'
E       - WSTestPkg/module_2/X64/TestFile.inf
E       ?          ^        ^   ^
E       + WSTestPkg\module_2\X64\TestFile.inf
```

### Expected behavior

The function should ensure that it handles input in a predictable and well-defined manner.

### Execution Environment

_No response_

### Pip packages

_No response_

### Additional context

_No response_

Javagedes:
Updates path_utilities to use pathlib rather than os.path functionality as it abstracts away posix vs windows path separators.

Continues to provides self.WorkspacePath and self.PackagePath as a string and list of strings respectively, but internally they are held as a pathlib.Path object and internal functionality uses it as such.

Resolves an issue where GetEdk2RelativePathFromAbsolutePath could return a Non-Posix path.

closes #207 

Also making initial updates in preparation for deprecating mws (multiple_workspace.py in edk2-pytool-extensions) found in https://github.com/tianocore/edk2-pytool-extensions/issues/446

### DIFF
--- a/edk2toollib/uefi/edk2/path_utilities.py
+++ b/edk2toollib/uefi/edk2/path_utilities.py
@@ -5,19 +5,29 @@
 #
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 ##
-"""Code to help convert Edk2, absolute, and relative file paths."""
+r"""A module for managing Edk2 file paths agnostic to OS path separators ("/" vs "\").
+
+This module converts all windows style paths to Posix file paths internally, but will return
+the OS specific path with the exception of of any function that returns an Edk2 style path,
+which will always return Posix form.
+"""
 import errno
-import fnmatch
 import logging
 import os
 from pathlib import Path
-from typing import Iterable
+from typing import Iterable, Optional
 
 
 class Edk2Path(object):
     """Represents edk2 file paths.
 
     Class that helps perform path operations within an EDK workspace.
+
+    Attributes:
+        WorkspacePath (str): Absolute path to the workspace root.
+        PackagePathList (List[str]): List of absolute paths to a package.
+
+    Attributes are initialized by the constructor and are read-only.
 
     !!! warning
         Edk2Path performs expensive packages path and package validation when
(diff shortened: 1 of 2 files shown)


edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -36,54 +46,52 @@ class Edk2Path(object):
 
     """
 
-    def __init__(self, ws: os.PathLike, package_path_list: Iterable[os.PathLike],
+    def __init__(self, ws: str, package_path_list: List[str],
```
makubacki, line 49: Does this need to be a list instead of a generic iterable? This particular argument is only being used with the iterator protocol in the for in loop.

makubacki (approved):

edk2toollib/uefi/edk2/path_utilities.py
Javagedes: How we consume it can be any iterable, but due to the comment below, we always return it as a list - I thought it would be confusing for a user to pass in a package path list as a tuple, then when they retrieve it, they get a list. to avoid the confusion, I changed it to say we only allow a lists. I can change it back and just make it a bit more clear in the description of Edk2Path that the package path list is manipulated.

```python
        # Other code is dependent the following types, so keep it that way:
        #   - self.PackagePathList: List[str]
        #   - self.WorkspacePath: str
```

makubacki: > How we consume it can be any iterable, but due to the comment below, we always return it as a list - I thought it would be confusing for a user to pass in a package path list as a tuple, then when they retrieve it, they get a list. to avoid the confusion, I changed it to say we only allow a lists. I can change it back and just make it a bit more clear in the description of Edk2Path that the package path list is manipulated.
> 
> ```python
>         # Other code is dependent the following types, so keep it that way:
>         #   - self.PackagePathList: List[str]
>         #   - self.WorkspacePath: str
> ```

I think it's fine to allow a generic iterable and return a list as long as the documentation (type hints included) indicate so.

Javagedes: Sounds good I've updated package_path_list type hint to be Iterable[str]. Thanks!

PR #378, opened 2023-08-03
path_utilities: allow path parts to be passed

Javagedes:
`GetEdk2RelativePathFromAbsolutePath` and `GetAbsolutePathOnThisSystemFromEdk2RelativePath` now allow the caller to provide the path parts as individual arguments instead of needing to join them together manually. Continues to support a single joined path argument.

This is to support the deprecation of [MultipleWorkspace](https://github.com/tianocore/edk2-pytool-extensions/blob/master/edk2toolext/environment/multiple_workspace.py) in edk2-pytool-extensions per https://github.com/tianocore/edk2-pytool-extensions/issues/446. With this change, users can now substitute `MultipleWorkspace.join(...)` for `Edk2Path.GetAbsolutePathOnThisSystemFromEdk2RelativePath(...)`.

Tests now include verifying the same output for manually joined paths and paths provided by parts when calling `GetEdk2RelativePathFromAbsolutePath` or `GetAbsolutePathOnThisSystemFromEdk2RelativePath`

This has been confirmed as a valid replacement on QemuQ35, QemuSbsaPkg, and other platforms.

## Breaking Change

This change is possibly breaking. Python allows providing kwargs as an arg (no keyword) so long as the number of parameters is correct. That is to say `Edk2Path.GetAbsolutePathOnThisSystemFromEdk2RelativePath(path, False)` is valid when technically it should be `Edk2Path.GetAbsolutePathOnThisSystemFromEdk2RelativePath(path, log_errors = False)`.

Now that `GetAbsolutePathOnThisSystemFromEdk2RelativePath` consumes all args, log_errors must be provided as a kwarg, otherwise it consumes the boolean and tries to append it to the path being created from the path parts.

### DIFF
--- a/edk2toollib/uefi/edk2/path_utilities.py
+++ b/edk2toollib/uefi/edk2/path_utilities.py
@@ -185,20 +185,22 @@ class Edk2Path(object):
         """List of package paths as strings."""
         return [str(p) for p in self._package_path_list]
 
-    def GetEdk2RelativePathFromAbsolutePath(self, abspath: str):
-        """Given an absolute path return a edk2 path relative to workspace or packagespath.
+    def GetEdk2RelativePathFromAbsolutePath(self, *abspath: str) -> str:
+        """Transforms an abolute path to an edk2 path relative to the workspace or a packages path.
 
         Args:
-            abspath: absolute path to a file or directory. Supports both Windows and Posix style paths
+            *abspath: absolute path to a file or directory. Can be the entire path or parts of the path provided
+                separately. Supports both Windows and Posix style paths
 
         Returns:
-            (str): POSIX-like relative path to workspace or packagespath
-            (None): abspath is none
+            (str): POSIX-like path relative to the workspace or packages path
+            (None): abspath is None
             (None): path is not valid
         """
-        if abspath is None:
+        if abspath == (None,):
             return None
-        abspath = Path(abspath.replace("\\", "/"))
+
+        abspath = Path(*[part.replace("\\", "/") for part in abspath])
 
         relpath = None
         found = False
@@ -234,11 +236,11 @@ class Edk2Path(object):
         self.logger.error(f'AbsolutePath: {abspath}')
         return None
 
-    def GetAbsolutePathOnThisSystemFromEdk2RelativePath(self, relpath: str, log_errors: Optional[bool]=True):
-        """Given a edk2 relative path return an absolute path to the file in this workspace.
+    def GetAbsolutePathOnThisSystemFromEdk2RelativePath(self, *relpath: str, log_errors: Optional[bool]=True) -> str:
+        """Given a relative path return an absolute path to the file in this workspace.
 
         Args:
-            relpath: Relative path to convert. Supports both Windows and Posix style paths.
+            *relpath: Relative path to convert. Can be the entire path or parts of the path provided separately
             log_errors: whether to log errors
 
         Returns:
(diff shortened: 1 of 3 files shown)


edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -185,20 +185,22 @@ def PackagePathList(self):
         """List of package paths as strings."""
         return [str(p) for p in self._package_path_list]
 
-    def GetEdk2RelativePathFromAbsolutePath(self, abspath: str):
-        """Given an absolute path return a edk2 path relative to workspace or packagespath.
+    def GetEdk2RelativePathFromAbsolutePath(self, *abspath: str) -> str:
+        """Transforms an abolute path to an edk2 path relative to the workspace or a packages path.
```
makubacki, line 189:
```suggestion
        """Transforms an absolute path to an edk2 path relative to the workspace or a packages path.
```

edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -185,20 +185,22 @@ def PackagePathList(self):
...
 
         Args:
-            abspath: absolute path to a file or directory. Supports both Windows and Posix style paths
+            *abspath: absolute path to a file or directory. Can be the entire path or parts of the path provided
+                separately. Supports both Windows and Posix style paths
```
makubacki, line 193: Since this is a new change to the interface, I think an example of calling with parts would be helpful to people considering using that.

edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -234,21 +236,22 @@ def GetEdk2RelativePathFromAbsolutePath(self, abspath: str):
...
         return None
 
-    def GetAbsolutePathOnThisSystemFromEdk2RelativePath(self, relpath: str, log_errors: Optional[bool]=True):
-        """Given a edk2 relative path return an absolute path to the file in this workspace.
+    def GetAbsolutePathOnThisSystemFromEdk2RelativePath(self, *relpath: str, log_errors: Optional[bool]=True) -> str:
+        """Given a relative path return an absolute path to the file in this workspace.
 
         Args:
-            relpath: Relative path to convert. Supports both Windows and Posix style paths.
+            *relpath: Relative path to convert. Can be the entire path or parts of the path provided separately
```
makubacki, line 243: Similar example here for consistency in documentation.

makubacki (approved):

edk2toollib/uefi/edk2/path_utilities.py
Javagedes: Added the following example and verified it display properly in the docs:

```python
abs_path = edk2path.GetAbsolutePathOnThisSystemFromEdk2RelativePath("MdePkg/Include")
abs_path = edk2path.GetAbsolutePathOnThisSystemFromEdk2RelativePath("MdePkg", "Include")
```

edk2toollib/uefi/edk2/path_utilities.py
Javagedes: Added the following example and verified it displayed properly in the docs:

```python
rel_path = edk2path.GetEdk2RelativePathFromAbsolutePath("C:/Workspace/edk2/MdePkg/Include")
rel_path = edk2path.GetEdk2RelativePathFromAbsolutePath("C:/Workspace", "edk2", "MdePkg", "Include")
```

edk2toollib/uefi/edk2/path_utilities.py
Javagedes: Updated. Thanks for the catch!

PR #379, opened 2023-08-03
ansi_handler: support all string representable objects in LogRecord message

### LINKED ISSUE #143: Pytool injecting extra list elements during logging
When logging level is higher than warning level, the ANSI handler from pytool will inject extra element to the message. https://github.com/tianocore/edk2-pytool-library/blob/0eac598e871e8f74068ac00089bf39f09d1264eb/edk2toollib/log/ansi_handler.py#L201

This is a problem when the incoming message is a list as is, since the input list will be updated by this statement and the next consumer using the same list will have an updated list with ANSI information inside.

Javagedes:
Previously, the ansi_handler only accepted strings in the logging record message due to a bug in how the message was retrieved from the record. This change fixes this, allowing any object that `str()` works on to be used as a message in the logging record.

This fix works by converting the message to a str (similar to `LogRecord.getMessage()`) then appending the color code. We cannot call `.getMessage()` directly as that will insert any args into the string, which causes a break later when `.getMessage()` is called again by the logging library (An exception is raised due to attempting to insert an argument into a string without a corresponding `%` placeholder).

closes #143

### DIFF
--- a/edk2toollib/log/ansi_handler.py
+++ b/edk2toollib/log/ansi_handler.py
@@ -203,7 +203,7 @@ class ColoredFormatter(logging.Formatter):
     def format(self, record):
         """Formats the given record and returns it."""
         levelname = record.levelname
-        message = record.getMessage()
+        org_message = record.msg
 
         if not self.use_azure and levelname in ColoredFormatter.COLORS:
             # just color the level name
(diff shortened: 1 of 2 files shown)


edk2toollib/log/ansi_handler.py
makubacki, line 1: If I understand correctly, I think the title would be a little easier to relate to the changes if something like "ansi_handler: support all string representable objects in record message".

makubacki (approved):

edk2toollib/log/ansi_handler.py
Javagedes: Updated title. I agree it will make more sense that way if someone reviews the git history.

PR #384, opened 2023-08-09
gitignore_parser: merge upstream changes

Javagedes:
Merges upstream changes of gitignore_parser.

The upstream changes were needed to add support for ignoring files with no extension via the following rule triple:

```txt
*     [Ignore all]
!*.*  [Un-ignore all files with extensions]
!*/   [Un-ignore all directories]
```

Puts us at [133bd62562622be096f495fbca7b37a1faac3ab7](https://github.com/mherrmann/gitignore_parser/commit/133bd62562622be096f495fbca7b37a1faac3ab7)

### DIFF
--- a/edk2toollib/gitignore_parser.py
+++ b/edk2toollib/gitignore_parser.py
@@ -123,24 +123,31 @@ def rule_from_pattern(pattern, base_path=None, source=None):
     anchored = '/' in pattern[:-1]
     if pattern[0] == '/':
         pattern = pattern[1:]
-    if pattern[0] == '*' and pattern[1] == '*':
+    if pattern[0] == '*' and len(pattern) >= 2 and pattern[1] == '*':
         pattern = pattern[2:]
         anchored = False
     if pattern[0] == '/':
         pattern = pattern[1:]
     if pattern[-1] == '/':
         pattern = pattern[:-1]
+    # patterns with leading hashes are escaped with a backslash in front, unescape it
+    if pattern[0] == '\\' and pattern[1] == '#':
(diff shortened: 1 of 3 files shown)


edk2toollib/gitignore_parser.py
```diff
@@ -123,24 +123,31 @@ def rule_from_pattern(pattern, base_path=None, source=None):
...
+    if pattern[0] == '*' and len(pattern) >= 2 and pattern[1] == '*':
         pattern = pattern[2:]
         anchored = False
     if pattern[0] == '/':
         pattern = pattern[1:]
     if pattern[-1] == '/':
         pattern = pattern[:-1]
+    # patterns with leading hashes are escaped with a backslash in front, unescape it
+    if pattern[0] == '\\' and pattern[1] == '#':
+        pattern = pattern[1:]
```
makubacki, line 135: For some of the lines being flagged as not covered. Can you add simple tests for those?

edk2toollib/gitignore_parser.py
makubacki, line 1: Can you include the SHA in the PR description of the point this is in sync with the upstream changes for future reference?

makubacki (approved):

edk2toollib/gitignore_parser.py
Javagedes: Updated the sha in the file and also in the PR description

edk2toollib/gitignore_parser.py
Javagedes: No problem, I've added additional tests to cover that, and another area that is uncovered after updates.

PR #393, opened 2023-08-21
database: multiple bugfixes

Javagedes:
This PR contains the following database changes:

### [bug] Account for library instances that do not exist in the DSC

The instanced_inf_table generator did not account for the fact that
library class instances can choose to only support phases.

1. Updates the DSC parser's ScopedLibraryDict to store a list of
instances that match the scope, instead of the first instance it finds.

2. The instanced_inf_table generator now checks the list of instances (
    in order of appearance in the DSC) to find the first instance that
    matches the scope, but also supports the specific phase.

### [bug] Account for library instances that support only a subset of phases.

Due to platforms performing complex includes in their DSCs, There is a
scenario in which a library instance references a library class that
does not exist. This is not an error as the library class instance may
not be consumed by a component in the platform, so the missing
dependency is acceptable.

This commit adds the ability to handle this scenario, and will not raise
an error for a missing library instance.

### [enhancement] Include library class and library instance in "LIBRARIES_USED" column of the instanced_inf table

The LIBRARIES_USED column in the instanced_inf table only contained
the the edk2 relative path to the library instance. This made it
impossible to determine (with 100% accuracy) which library class the
particular library instance was representing.

This commit changes the "LIBRARIES_USED" column to contain a list of
(lib_class, lib_instance) tuples rather than a list of lib_instances.
Please review commit messages in each for a more detailed description.

### [enhancement] Include "LIBRARY_CLASS" column in instanced_inf table

Previously, the library class name for a library INF was not included
in the table. This commit adds that information to the table as a row
to make it easier to find the library class name for a library INF.

### [bug] Better handling of RuleOverride in instanced_fv_table

When parsing an FDF file, an INF can be prepending with a RuleOverride
directive. The previous functionality to ignore this directive and
retrive the actual INF value could not handle any spaces in the
directive (such as RuleOverride = ACPITABLE). This commit expands the 
ability to handle spaces in the directive.


edk2toollib/uefi/edk2/parsers/inf_parser.py
```diff
@@ -12,7 +12,7 @@
 from edk2toollib.uefi.edk2.parsers.base_parser import HashFileParser
 
 AllPhases = ["SEC", "PEIM", "PEI_CORE", "DXE_DRIVER", "DXE_CORE", "DXE_RUNTIME_DRIVER", "UEFI_DRIVER",
-             "SMM_CORE", "DXE_SMM_DRIVER", "UEFI_APPLICATION"]
+             "SMM_CORE", "DXE_SMM_DRIVER", "UEFI_APPLICATION", "MM_STANDALONE", "MM_CORE_STANDALONE"]
```
cfernald, line 15: Should we include HOST_APPLICATION?

edk2toollib/database/tables/instanced_fv_table.py
```diff
@@ -27,6 +28,9 @@ class InstancedFvTable(TableGenerator):
     |------------------------------------------------------|
     ```
     """  # noqa: E501
+
+    RULEOVERRIDE = re.compile(r'RuleOverride\s*=.+\s+(.*\.inf)', re.IGNORECASE)
```
makubacki, line 32: `.*\.inf` was intentionally allowing `.inf` to match, right? As opposed to `.+\.inf` which would require at least one character in the name. I think allowing only the extension to match is fine, but it could also be a mistake.

cfernald (approved):

Pipfile
makubacki, line 1: Was this file meant to be in this commit?

makubacki (approved):

edk2toollib/uefi/edk2/parsers/inf_parser.py
Javagedes: Absolutely. I'll add it in. This list was fairly outdated so if there are more then just HOST_APPLICATION, let me know.

edk2toollib/database/tables/instanced_fv_table.py
Javagedes: I agree - it's more appropriate to require at least a single character before `.inf`. I will update. :)

Pipfile
Javagedes: Yes this was a mistake, I will remove.

Javagedes: This was a mistake, but I'm not doing a rebase and FF, so I just removed it in a later commit and called it a day.

If you would prefer a rebase & FF, let me know - I will fix this

makubacki: Not an issue if you're squashing.

edk2toollib/uefi/edk2/parsers/inf_parser.py
Javagedes: Added. Thanks for the review!

edk2toollib/database/tables/instanced_fv_table.py
Javagedes: Updated! Thanks for the review!

PR #399, opened 2023-08-28
base_parser: support relative path include files

### LINKED ISSUE #394: [Bug]: `base_parser.TargetFilePath` not reverted when exiting an !include parse
### Contact Details

_No response_

### Describe the Bug

Per [section 2.2.5](https://tianocore-docs.github.io/edk2-DscSpecification/release-1.28/2_dsc_overview/22_build_description_file_format.html#22-build-description-file-format) of the DSC specification, when searching for an !include filepath, the first location to check is relative to the directory containing the currently parsed file. 

edk2-pytool-library emulates this functionality by storing the currently parsed file in `TargetFilePath` so that if we need to find another include, we are looking in the correct relative directory. The bug, is that we do not revert `TargetFilePath` to the previous file when we are done parsing an included file. This means that searches for additional includes could fail as the relative directory we look for is incorrect.

#### Example

Folder layout we are parsing
``` cmd
src
└─ Platforms
  └─ PlatformPkg
    ├─ PlatformPkg.dsc
    └─ includes
      ├─ PCDs1.dsc.inc
      └─ PCDs2.dsc.inc   
```

Abbreviated DSC we are parsing
```
[PcdsFixedAtBuild]
gPlatformPkgTokenSpaceGuid.PcdSomethig
!include includes/PCDs1.dsc.inc
!include includes/PCDs2.dsc.inc
```

When we start parsing PlatformPkg.dsc, `TargetFilePath` is `~/src/Platforms/PlatformPkg/PlatformPkg.dsc`.

When we start parsing the first include, the relative import is a success because the `~src/Platforms/PlatformPkg/` + `includes/PCDs1.dsc.inc` exists. We then update TargetFilePath to `~/src/Platforms/PlatformPkg/includes/PCDs1.dsc.inc`.

When we start parsing the second include, we fail because `TargetFilePath` was not reverted back to `~/src/Platforms/PlatformPkg/PlatformPkg.dsc`. Since it remains the same, we think the file path should be `~src/Platforms/PlatformPkg/includes/includes/PCDs2.dsc.inc`, which is wrong.

### Reproduction steps

Update your two dscs to use two relative path includes (not macro defined absolute paths or edk2 package path relative paths) and the parsing will fail.

```
!include relative/myfile1.dsc
!include relative/myfile2.dsc
!include relative/myfile3.dsc
```

### Expected behavior

Parsing should correctly update TargetFilePath and relative includes should not fail.

Python 3.9, Python 3.10, Python 3.11

### Execution Environment

_No response_

### Pip packages

_No response_

### Additional context

_No response_

Javagedes:
This commit adds support for including multiple relative paths from the same directory by storing a stack of all included files that have not finished being parsed, and the number of lines that still need to be parsed in each file.

Previously, only includes of nested relative paths were supported as the currently parsed file was not reverted to the previous file when the parsing of an included file was completed. This led to further includes needing the be relative to the latest included file.

closes #394

### DIFF
--- a/edk2toollib/uefi/edk2/parsers/base_parser.py
+++ b/edk2toollib/uefi/edk2/parsers/base_parser.py
@@ -42,7 +42,7 @@ class BaseParser(object):
         self.PPs = []
         self._Edk2PathUtil = None
         self.TargetFilePath = None  # the abs path of the target file
-        self.FilePathStack = []  # a stack containing a tuple of a file, and lines left to parse in that file
+        self.FilePathStack = []  # a stack containing a list of size 2: [filepath, line_count]
         self.ParsedFiles = set()
         self.CurrentLine = -1
         self._MacroNotDefinedValue = "0"  # value to used for undefined macro
@@ -137,7 +137,7 @@ class BaseParser(object):
 
     def PushTargetFile(self, abs_path, line_count):
         """Adds a target file to the stack."""
-        self.FilePathStack.append((abs_path, line_count))
+        self.FilePathStack.append([abs_path, line_count])
         self.ParsedFiles.add(abs_path)
 
     def DecrementLinesParsed(self) -> bool:
@@ -148,9 +148,9 @@ class BaseParser(object):
         """
         if not self.FilePathStack:
             return False
-        (abs_path, line_count) = self.FilePathStack[-1]
+        line_count = self.FilePathStack[-1][1]
         if line_count - 1 > 0:
-            self.FilePathStack[-1] = (abs_path, line_count - 1)
+            self.FilePathStack[-1][1] -= 1
             return True
 
         self.FilePathStack.pop()
--- a/edk2toollib/uefi/edk2/parsers/base_parser.py
+++ b/edk2toollib/uefi/edk2/parsers/base_parser.py
@@ -42,6 +42,8 @@ class BaseParser(object):
         self.PPs = []
         self._Edk2PathUtil = None
         self.TargetFilePath = None  # the abs path of the target file
+        self.FilePathStack = []  # a stack containing a tuple of a file, and lines left to parse in that file
+        self.ParsedFiles = set()
         self.CurrentLine = -1
         self._MacroNotDefinedValue = "0"  # value to used for undefined macro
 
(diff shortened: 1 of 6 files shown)


Flickdm (approved):

edk2toollib/uefi/edk2/parsers/base_parser.py
```diff
@@ -42,6 +42,8 @@ def __init__(self, log="BaseParser"):
         self.PPs = []
         self._Edk2PathUtil = None
         self.TargetFilePath = None  # the abs path of the target file
+        self.FilePathStack = []  # a stack containing a list of size 2: [filepath, line_count]
```
Flickdm, line 45: NIT: 

Since you're likely dealing with a large set of data, the following solution probably doesn't make a lot of sense performance wise but if you wanted to improve readability of the code you could do something like a namedtuple.

However the object instantiation might be too much for a large dataset

```python

from collections import namedtuple

# Define the namedtuple type
FilePathRecord = namedtuple('FilePathRecord', ['filepath', 'line_count'])

self.FilePathStack = []  # a stack containing FilePathRecord instances

self.FilePathStack.append(FilePathRecord('my_file_path', 100))

self.FilePathStack[0].file_path
```

edk2toollib/uefi/edk2/parsers/base_parser.py
```diff
@@ -133,6 +135,27 @@ def SetInputVars(self, inputdict):
...
+        """Decrements line count for the current target file by one.
+
+        Returns:
+            (bool): True if there are still lines to parse, False otherwise.
+        """
+        if not self.FilePathStack:
+            return False
+        line_count = self.FilePathStack[-1][1]
+        if line_count - 1 > 0:
+            self.FilePathStack[-1][1] -= 1
```
Flickdm, line 153: NIT:
```python
        updated_line_count = self.FilePathStack[-1][1] - 1
        if updated_line_count > 0:
            self.FilePathStack[-1][1] -= updated_line_count
```

edk2toollib/uefi/edk2/parsers/dsc_parser.py
```diff
@@ -92,10 +92,10 @@ def __ParseLine(self, Line, file_name=None, lineno=None):
             if sp is None:
                 raise FileNotFoundError(include_file)
             self.Logger.debug("Opening Include File %s" % sp)
-            self._PushTargetFile(sp)
             lf = open(sp, "r")
```
Flickdm, line 95: NIT: I know you didn't write this but if we switch to with we can drop lf.close()

Flickdm (approved):

PR #406, opened 2023-09-06
edk2toollib/database: Switch to sqlite3

Javagedes:
Switch to using sqlite3 rather than TinyDB as this is a true database that supports SQL queries and other expected functionality that TinyDB does not.

Additionally:

- Removes all queries as they were just a bridge due to the fact that TinyDB did not support SQL queries and in particular, joins.
- Adds a Package Table, which can contain generic information about the package (right now, only the containing repository)


edk2toollib/database/edk2_db.py
```diff
@@ -7,107 +7,74 @@
...
-    1. File Read/Write: A database will be loaded or created at the specified path. Any changes made will be written
-       to the database file. This is the slowest of the three modes. Specify with Edk2DB.FILE_RW
+    Attributes:
+        connection (sqlite3.Connection): The connection to the database
 
-    2. File Read Only: A database will be loaded at the specific path. Attempting to change the database will result
-       in an error. This is the middle of the three in terms of performance. Specify with Edk2DB.FILE_RO
+    !!! note
+        Edk2DB provides a table called `junction` that can be used to make associations between tables. It has the
+        following schema: `table1, key1, table2, key2`.
```
Javagedes, line 46:
```suggestion
        following schema: `env_id, table1, key1, table2, key2`.
```

edk2toollib/database/tables/package_table.py
```diff
@@ -0,0 +1,59 @@
+# @file repository_table.py
```
makubacki, line 1:
```suggestion
# @file package_table.py
```

edk2toollib/database/tables/environment_table.py
```diff
@@ -6,46 +6,54 @@
...
+        dtime = datetime.datetime.now()
+
+        try:
+            version = git.Repo(pathobj.WorkspacePath).head.commit.hexsha
+        except git.InvalidGitRepositoryError:
+            version = "UNKNOWN"
+
+        # Insert into environment table
+        entry = (id,str(dtime),version,)
+        db_cursor.execute("INSERT INTO environment (id, date,version) VALUES (?, ?, ?)", entry)
```
Javagedes, line 55:
```suggestion
        db_cursor.execute("INSERT INTO environment (id, date, version) VALUES (?, ?, ?)", entry)
```

makubacki (approved):

edk2toollib/database/__init__.py
makubacki, line 1: I suggest making the PR title somthing like `edk2toollib/database: Switch to sqlite3`.

edk2toollib/database/tables/package_table.py
Javagedes: Updated. Thank you!

edk2toollib/database/__init__.py
Javagedes: Updated. Thank you!

PR #419, opened 2023-09-29
edk2toollib/uefi/edk2/parsers: enhancements

Javagedes:
Improves the following parsers in the following ways:

`base_parser.py`: Properly uses ivalue (int value) to check if the string representation of a value was correctly converted to an integer ('1' to 1, '0x0' to 0, etc) when evaluating conditionals

`dsc_parser.py`: register pcds when performing the initial parse responsible for finding define statements. This is necessary to evaluate conditionals that use pcds.

### DIFF
--- a/edk2toollib/uefi/edk2/parsers/base_parser.py
+++ b/edk2toollib/uefi/edk2/parsers/base_parser.py
@@ -240,12 +240,12 @@ class BaseParser(object):
             return (ivalue != ivalue2) and (value != value2)
 
         # check to make sure we only have digits from here on out
-        if not isinstance(value, int) and not str.isdigit(value):
+        if not isinstance(ivalue, int) and not str.isdigit(value):
             self.Logger.error(f"{self.__class__}: Unknown value: {value} {ivalue.__class__}")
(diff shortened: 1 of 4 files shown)


apop5 (approved):

edk2toollib/uefi/edk2/parsers/base_parser.py
```diff
@@ -240,12 +240,17 @@ def ComputeResult(self, value, cond, value2):
             return (ivalue != ivalue2) and (value != value2)
 
         # check to make sure we only have digits from here on out
```
makubacki, line 242: It seems this line should go after this block and before the `isdigit()` checks below?

makubacki (approved):

edk2toollib/uefi/edk2/parsers/base_parser.py
Javagedes: All three of the checks below this line are to insure we only have digits from here on out (verifying value is not a string representation of true / false, and isinstance / isdigit checks).

I can't rely soley on isinstance and isdigit because `self.ConvertToInt` converts the any TRUE / FALSE strings to 1 / 0, so we have to make sure the origin value was not a TRUE/FALSE also.

PR #423, opened 2023-10-03
[REBASE&FF] Multiple table / parser updates

Javagedes:
Contains multiple commits to improve `instanced_table`, `instanced_inf_table`, and `dsc_parser`.

## Commit 1: inf_table.py: resolve dot-dot paths in source files

Updates inf_table.py to resolve dot-dot paths in source files before inserting it into the database to ensure the path is a valid edk2 package path relative path. Source files in the database are now edk2 package relative rather than inf relative.

## Commit 2: instanced_inf_table.py: Filter on TARGET_ARCH

The dsc_parser is architecture agnostic and resolves the dsc file for all architectures it supports. This means the instanced_inf_table must filter the results to only the architectures specified by the Settings file.

## Commit 3: dsc_parser.py: use SUPPORTED_ARCHITECTURES

The dsc_parser relied on TARGET_ARCH when creating scoped libraries and components, but it should rely on SUPPORTED_ARCHITECTURES which provides information true to the DSC.

## Commit 4: inf parsing: Add support for scoped source files

Adds additional parsing to the inf_parser to support scoped source files
which can be retrieved via the function get_sources(arch_list), which
will return all sources used for a given arch(s). Re-writes the logic
that does the same for scoped libraries to make it more extensible, but
no user facing changes.

Updates instanced_inf_table to use the new function to get the sources.

## Commit 5: inf_parser: Add ReplaceVariable support

Previously, the INF parser did not support detecting DEFINE variables or
replacing $(variable) with the value of the variable, similar to other
parsers. This change adds support for this feature, and assocaited tests
to ensure it works as expected.

### DIFF
--- a/edk2toollib/database/tables/instanced_inf_table.py
+++ b/edk2toollib/database/tables/instanced_inf_table.py
@@ -206,6 +206,7 @@ class InstancedInfTable(TableGenerator):
         visited.append(inf)
         library_instance_list = []
         library_class_list = []
+        arch = scope.split(".")[0].upper()
 
         #
         # 0. Use the existing parser to parse the INF file. This parser parses an INF as an independent file
@@ -217,7 +218,7 @@ class InstancedInfTable(TableGenerator):
         # 1. Convert all libraries to their actual instances for this component. This takes into account
         #    any overrides for this component
         #
-        for lib in infp.get_libraries(self.arch):
+        for lib in infp.get_libraries([arch]):
             lib = lib.split(" ")[0]
             library_instance_list.append(self._lib_to_instance(lib.lower(), scope, library_dict, override_dict))
             library_class_list.append(lib)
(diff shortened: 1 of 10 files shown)


edk2toollib/database/tables/instanced_inf_table.py
```diff
@@ -164,7 +164,12 @@ def _build_inf_table(self, dscp: DscP):
         """
         inf_entries = []
         for (inf, scope, overrides) in dscp.Components:
-            logging.debug(f"Parsing Component: [{inf}]")
+            # Ignore components built with an architecture that is not in TARGET_ARCH
+            arch = scope.split(".")[-1].upper()
```
cfernald, line 168: For my education, why the switch from forward index like with the removed `"ARCH": scope.split(".")[0].upper(),` to backward index?

cfernald (approved):

edk2toollib/database/tables/instanced_inf_table.py
Javagedes: This was actually a mistake on my part. Thank you for catching this. It has been switched back to 0

PR #430, opened 2023-10-09
path_utilities: Remove Nested package exceptions

### LINKED ISSUE #200: [Bug]: Remove PYTOOL_TEMPORARILY_IGNORE_NESTED_EDK_PACKAGES
### Contact Details

_No response_

### Describe the Bug

This environment variable should be removed as it was added as a temporary measure for downstream consumers to not have to fix their nested packages right away.

PYTOOL_TEMPORARILY_IGNORE_NESTED_EDK_PACKAGES

### Reproduction steps

Review https://github.com/tianocore/edk2-pytool-library/pull/199

### Expected behavior

The variable and related code are removed. A release message is provided describing the removal when it is done.

### Execution Environment

_No response_

### Pip packages

_No response_

### Additional context

_No response_

Javagedes:
Removes Nested package exceptions from path_utilities.py in favor of a warning.  This is being done as the exception has served its purpose and those willing to update their nested packages have had a year to do so. Any nested packages will now only be a warning logged at logging.DEBUG and issues with converting from an absolute path to an edk2 package path relative path will be accepted.

A new CI Check has been added so that packages managed by that specific repository will still be checked by nested packages, but nested packages outside of the control of that specific repository will not be checked and thus not result in an exception being raised.

closes #200

### DIFF
--- a/edk2toollib/uefi/edk2/path_utilities.py
+++ b/edk2toollib/uefi/edk2/path_utilities.py
@@ -105,6 +105,15 @@ class Edk2Path(object):
         #    is relative (nested) to each other.
         # 3. Raise an Exception if two packages are found to be nested.
         #
+        if "PYTOOL_TEMPORARILY_IGNORE_NESTED_EDK_PACKAGES" in os.environ:
+            warning = "PYTOOL_TEMPORARILY_IGNORE_NESTED_EDK_PACKAGES is no longer used by edk2-pytool-library, but is "\
+                      "detected in your environment. Please remove this environment variable."
+            self.Logger.log(logging.WARNING, warning)
+        if "PYTOOL_IGNORE_KNOWN_BAD_NESTED_PACKAGES" == os.environ:
+            warning = "PYTOOL_IGNORE_KNOWN_BAD_NESTED_PACKAGES is no longer used by edk2-pytool-library, but is " \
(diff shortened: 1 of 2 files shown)


edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -120,60 +110,18 @@ def __init__(self, ws: str, package_path_list: Iterable[str],
...
 
-        # Note: The ability to ignore this function raising an exception on
-        #       nested packages is temporary. Do not plan on this variable
-        #       being available long-term and try to resolve the nested
-        #       packages problem right away.
-        #
-        # Removal is tracked in the following GitHub issue:
-        # https://github.com/tianocore/edk2-pytool-library/issues/200
-        ignore_nested_packages = False
-        if "PYTOOL_TEMPORARILY_IGNORE_NESTED_EDK_PACKAGES" in os.environ and \
```
makubacki, line 131: If the variables are still being set after the change, it might be worth adding a warning message that the variables are now deprecated and ineffective. To help clean up consumer code and reduce confusion about what the variables might do if people see them still sitting around.

Javagedes: I have added a warning that they are no longer used, and to remove them from your environment.

Thanks!

PR #437, opened 2023-10-23
gitignore_parser.py: Multiple Updates

Javagedes:
This squash commit merges in changes from mherrmann/gitignore_parser for the following commits:

7956d03: remove unused variable whitespace_re
ffbfd79: remove regex flags m, s
d45a085: Fix pattern with slash in range
1040aa5: Fix pattern with leading exclamation marks
cdf80b7: Fix lack of implicit anchoring of patterns to direcotry separators
cdf80b7: Fix multi-astericks that fall outside of the special cases
6abc776: Fix "a/**/b" matching "a/bb"
721f804: do not resolve symlinks

This is specifically necessary for **721f804: do not resolve symlinks** which was causing the GuidCheck plugin to fail for Edk2 as it has a symbolic link: https://github.com/tianocore/edk2/blob/master/EmulatorPkg/Unix/Host/X11IncludeHack

### DIFF
--- a/tests.unit/parsers/test_gitingore_parser.py
+++ b/tests.unit/parsers/test_gitingore_parser.py
@@ -89,8 +89,10 @@ class GitIgnoreParserTest(unittest.TestCase):
 
             # Example line in a .gitignore
             # .DS_Store
-            self.assertTrue(rule_tester(root / "file.DS_Store"))
-            self.assertTrue(rule_tester(root / "T" / "file.DS_Store"))
+            self.assertTrue(rule_tester(root / ".DS_Store"))
+            self.assertTrue(rule_tester(root / "T" / ".DS_Store"))
+            self.assertFalse(rule_tester(root / "file.DS_Store"))
+            self.assertFalse(rule_tester(root / "T" / "file.DS_Store"))
 
             # Test a rule which specifies an exception for a previous rule is
             # correctly filtered.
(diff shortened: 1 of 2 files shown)


tests.unit/parsers/test_gitingore_parser.py
```diff
@@ -175,32 +178,144 @@ def test_pound_in_filename(tmp_path):
...
+    assert rule_tester('/home/tmp/partiallyignoredspace2   ') is False
+    assert rule_tester('/home/tmp/partiallyignoredspace2 ') is False
+    assert rule_tester('/home/tmp/partiallyignoredspace2') is False
+    assert rule_tester('/home/tmp/notignoredspace ') is True
+    assert rule_tester('/home/tmp/notignoredspace') is False
+    assert rule_tester('/home/tmp/notignoredmultiplespace   ') is True
+    assert rule_tester('/home/tmp/notignoredmultiplespace') is False
+
+def test_slash_in_range_does_not_match_dirs(tmp_path):
+    """Tests that a slash in a range does not match directories."""
```
makubacki, line 209: Some of the test functions have docstrings and others do not. Was there a criteria in determining when to add them?

For consistency, a simple string for each would help set precedent for the future.

makubacki (approved):

tests.unit/parsers/test_gitingore_parser.py
Javagedes: It's mainly that tests that existed did not have documentation and tests that I've added over time did. I've updated all to have docstrings.

PR #456, opened 2023-11-20
base_parser: Update StripComment to consider quote groupings

Javagedes:
Previous update to StripComment would ignore any pound signs that were surrounded by quotes, but did not consider if the sign was inside two quotes or between two sets of two quotes (i.e. `" # "` vs `" 1 " # " 2 "`). This commit updates StripComment to check this scenario and not strip if the pound sign is inside two quotes.

This additionally adds support for ignoring escaped quotes.


apop5 (approved):
New comment parser works in the project.

makubacki (approved):

PR #541, opened 2024-04-02
Update OS version to 17134 to limit DDInstall sections

srilathasridharan:
Limit capsule install inf changes to OS versions 17134 or higher for a newer HDC requirement.
Original PR - https://github.com/tianocore/edk2-pytool-library/pull/431

"According to Zac Lockard for the InfVerif team, the new WHCP requirements went into force with the newest 26080 build which is why we're seeing this now. For reference, they link https://learn.microsoft.com/en-us/windows-hardware/drivers/devtest/inf-validation-errors-and-warnings#1199 and quote the page that says "You must fix all errors in order to pass driver tests in Hardware Dev Center."

It is the Device.DevFund.INF.Syntax requirement

This change was also pushed to all older versions of the OS due to this being a problem often in the ecosystem".

### DIFF
--- a/edk2toollib/windows/capsule/inf_generator.py
+++ b/edk2toollib/windows/capsule/inf_generator.py
@@ -52,9 +52,9 @@ PnpLockdown=1
 CatalogFile={Name}.cat
 
 [Manufacturer]
-%MfgName% = Firmware,NT{Arch}.10.0...16299
+%MfgName% = Firmware,NT{Arch}.10.0...17134
 
-[Firmware.NT{Arch}.10.0...16299]
+[Firmware.NT{Arch}.10.0...17134]
 %FirmwareDesc% = Firmware_Install,UEFI\RES_{{{EsrtGuid}}}
 
 [Firmware_Install.NT]
--- a/edk2toollib/windows/capsule/inf_generator.py
+++ b/edk2toollib/windows/capsule/inf_generator.py
@@ -17,6 +17,7 @@ import re
 import uuid
 from typing import Optional
 
+OS_BUILD_VERSION_DIRID13_SUPPORT = '10.0...17134'
 
 class InfSection(object):
     """Object representing an INF Section."""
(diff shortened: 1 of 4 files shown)


makubacki:
Do you have any links to documentation that state this requirement?

namiba-work:
> Do you have any links to documentation that state this requirement?

According to Zac Lockard for the InfVerif team, the new WHCP requirements went into force with the newest 26080 build which is why we're seeing this now. For reference, they link https://learn.microsoft.com/en-us/windows-hardware/drivers/devtest/inf-validation-errors-and-warnings#1199 and quote the page that says "You must fix all errors in order to pass driver tests in Hardware Dev Center."

edk2toollib/windows/capsule/inf_generator.py
```diff
@@ -331,6 +332,7 @@ def MakeInf(self, OutputInfFilePath: os.PathLike, FirmwareBinFileName: str, Roll
             EsrtGuid=self.EsrtGuid,
             Provider=self.Provider,
             MfgName=self.Manufacturer,
+            OsBuildVersion=OS_BUILD_VERSION_DIRID13_SUPPORT,
```
makubacki, line 335: This variable should similarly be called `OsBuildVersionDirId13Support` or something that allows it to carry that value forward 1:1 so it tracks the OS build version for that specific need.

makubacki (approved):

makubacki:
> > Do you have any links to documentation that state this requirement?
> 
> According to Zac Lockard for the InfVerif team, the new WHCP requirements went into force with the newest 26080 build which is why we're seeing this now. For reference, they link https://learn.microsoft.com/en-us/windows-hardware/drivers/devtest/inf-validation-errors-and-warnings#1199 and quote the page that says "You must fix all errors in order to pass driver tests in Hardware Dev Center."
> 
> It is the Device.DevFund.INF.Syntax requirement
> 
> This change was also pushed to all older versions of the OS due to this being a problem often in the ecosystem

@srilathasridharan, please include this info in the PR description so it is easy for people to find that come acrss this change in the future.

edk2toollib/windows/capsule/inf_generator.py
srilathasridharan: we could be restricting the version for other reasons too in the future, with another update, so I left that flexible, since I had an opportunity to do so in this file, while also indicating using the define name as to why is the version restriction in place.

PR #561, opened 2024-05-16
Correcting Incorrect Usage of EfiSignatureList

Flickdm:
An authenticated variable may contain only a single EfiSignatureList. However, they may contain multiple EfiSignatureLists otherwise known as EfiSignatureDatabase. 

This change corrects that misconception


Javagedes:
This looks fine, just needs to be updated to fix the unit tests

Javagedes (approved):

PR #577, opened 2024-06-06
Adding Uefi Variable Support.

apop5:
Adding a common Uefi Variable Access class for both windows and linux.

Class UefiVariable supports:
    GetUefiVar - Get a single Uefi Variable's data
    GetUefiAllVarNames - Get all Uefi variables
    SetUefiVar - Set (or delete) a single Uefi variable.

Multiple copies of this class are floating in repos, mainly to support tools. This is to consolidate support.

Verified DecodeUefiLog is able to use this instance to retrieve log on a physical system.

### DIFF
--- a/edk2toollib/os/uefivariablesupport.py
+++ b/edk2toollib/os/uefivariablesupport.py
@@ -106,13 +106,13 @@ class UefiVariable(object):
                 ]
             except Exception:
                 logging.warn(
-                    "G[S]etFirmwareEnvironmentVariableW function doesn't seem to exist"
+                    "Collecting windll Variable functions encountered an error."
                 )
         else:
             pass
 
     def GetUefiVar(self, name: str, guid: str) -> tuple[int, str]:
-        """_Retrieve Uefi Variable from the system.
+        """Get a Uefi Variable from the system.
 
         Args:
             name (str): String corresponding to the Unicode Name of the variable.
@@ -165,13 +165,13 @@ class UefiVariable(object):
 
         Returns:
             tuple[int, bytes]:
-            Integer is return status, 0 for error, non-zero for success.
-            Bytes are the packed structure for each variable.
-                struct _VARIABLE_NAME {
-                    ULONG NextEntryOffset;
-                    GUID VendorGuid;
-                    WCHAR Name[ANYSIZE_ARRAY];
-                }
+                Integer is return status, 0 for error, non-zero for success.
+                Bytes are the packed structure for each variable.
+                    struct _VARIABLE_NAME {
+                        ULONG NextEntryOffset;
+                        GUID VendorGuid;
+                        WCHAR Name[ANYSIZE_ARRAY];
+                    }
         """
         status = 0
         if os.name == 'nt':
@@ -254,13 +254,13 @@ class UefiVariable(object):
             return (status, efi_var_names)
 
     def SetUefiVar(self, name: str, guid: str, var: str = None, attrs: int = None) -> int:
-        """_Set a Uefi Variable into the system.
+        """Set a Uefi Variable into the system.
 
         Args:
             name (str): Unicode name of variable to set
             guid (str): Guid to use when setting the variable
-            var (bytes, optional): Bytes to set to the variable. Defaults to None.
-            attrs (_type_, optional): Attributes to use when setting the variable. Defaults to None.
+            var (Optional[bytes]): Bytes to set to the variable. Defaults to None.
+            attrs (Optional[int]): Attributes to use when setting the variable. Defaults to None.
 
         Returns:
             int: 0 for a failure, non-zero for success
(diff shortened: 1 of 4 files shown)


edk2toollib/os/uefivariablesupport.py
```diff
@@ -0,0 +1,331 @@
...
+                ]
+            except Exception:
+                logging.warn(
+                    "G[S]etFirmwareEnvironmentVariableW function doesn't seem to exist"
+                )
+        else:
+            pass
+
+    def GetUefiVar(self, name: str, guid: str) -> tuple[int, str]:
+        """_Retrieve Uefi Variable from the system.
```
Javagedes, line 115: nit: don't need the "_"

edk2toollib/os/uefivariablesupport.py
```diff
@@ -0,0 +1,331 @@
...
+
+    def GetUefiAllVarNames(self) -> tuple[int, bytes]:
+        """Get all Uefi Variables in the system, and return a byte packed byte string.
+
+        Raises:
+            Exception: Returned variable could not be parsed.
+
+        Returns:
+            tuple[int, bytes]:
+            Integer is return status, 0 for error, non-zero for success.
```
Javagedes, line 168: The documentation after `tuple[int, bytes]: needs to be indented. That will fix the ruff error.

edk2toollib/os/uefivariablesupport.py
```diff
@@ -0,0 +1,331 @@
...
+                offset += 16
+
+                # Name
+                struct.pack_into(f'={name_len}s', efi_var_names, offset, name)
+                offset += name_len
+
+            return (status, efi_var_names)
+
+    def SetUefiVar(self, name: str, guid: str, var: str = None, attrs: int = None) -> int:
+        """_Set a Uefi Variable into the system.
```
Javagedes, line 257: nit: don't need the "_"

edk2toollib/os/uefivariablesupport.py
```diff
@@ -0,0 +1,331 @@
...
+
+        Args:
+            name (str): Unicode name of variable to set
+            guid (str): Guid to use when setting the variable
+            var (bytes, optional): Bytes to set to the variable. Defaults to None.
```
Javagedes, line 262: `Optional[bytes]`

edk2toollib/os/uefivariablesupport.py
```diff
@@ -0,0 +1,331 @@
...
+            attrs (_type_, optional): Attributes to use when setting the variable. Defaults to None.
```
Javagedes, line 263: `Optional[int]`

edk2toollib/os/uefivariablesupport.py
```diff
@@ -0,0 +1,331 @@
...
+                self._SetFirmwareEnvironmentVariableEx.argtypes = [
+                    c_wchar_p,
+                    c_wchar_p,
+                    c_void_p,
+                    c_int,
+                    c_int,
+                ]
+            except Exception:
+                logging.warn(
+                    "G[S]etFirmwareEnvironmentVariableW function doesn't seem to exist"
```
Javagedes, line 109: Was this `G[S]et` on purpose or a typo?

apop5: I think it was intentional.

The __init__ function is trying to collect the following functions:
GetFirmwareEnvironmentVariableW
NtEnumerateSystemEnvironmentValuesEx
SetFirmwareEnvironmentVariableW
SetFirmwareEnvironmentVariableExW

Its the same error message for any exception thrown. I agree it's confusing, so I will clarify it to something more generic like "Collecting windll Variable functions encountered an error."

edk2toollib/os/uefivariablesupport.py
apop5: Fixed in latest.

edk2toollib/os/uefivariablesupport.py
apop5: Fixed in latest.

edk2toollib/os/uefivariablesupport.py
apop5: Fixed in latest.

edk2toollib/os/uefivariablesupport.py
apop5: Fixed in latest.

edk2toollib/os/uefivariablesupport.py
apop5: Fixed in latest.

edk2toollib/os/uefivariablesupport.py
apop5: Fixed in latest.

Javagedes (approved):

PR #600, opened 2024-07-19
Fix PCD parsing for PCDs that have a single quote in them

NishanthSanjeevi:
The PCD parsing regex pattern failed for PCDs that had a single quote

For example, 
gEfiMdePkgTokenSpaceGuid.PcdSingleQuote|"eng'fraengfra"|VOID*|0x0000002f
CRITICAL - EXCEPTION: Too few parts: ['PcdSingleQuote|"eng\'fraengfra"', 'VOID*', '0x0000002f']

The new pattern should allow the usage of single quotes

### DIFF
--- a/edk2toollib/uefi/edk2/parsers/dec_parser.py
+++ b/edk2toollib/uefi/edk2/parsers/dec_parser.py
@@ -127,7 +127,7 @@ class PcdDeclarationEntry():
         self.token_space_name = sp[0].strip()
 
         # Regular expression pattern to match the symbol '|' that is not inside quotes
-        pattern = r'\|(?=(?:[^\'"]*[\'"][^\'"]*[\'"])*[^\'"]*$)'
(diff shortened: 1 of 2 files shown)


Javagedes:
@NishanthSanjeevi Please add a test or two for this PR that test the new changes and supported logic. The test(s) would make sense here:

https://github.com/tianocore/edk2-pytool-library/blob/36dd4436ce4a1237a41e9964770540a173abaf3d/tests.unit/parsers/test_dec_parser.py#L81

Javagedes (changes requested):
Please add a test or two for this PR that test the new changes and supported logic. The test(s) would make sense here:

https://github.com/tianocore/edk2-pytool-library/blob/36dd4436ce4a1237a41e9964770540a173abaf3d/tests.unit/parsers/test_dec_parser.py#L81

Javagedes (approved):

PR #644, opened 2024-09-30
gitignore_parser.py: Handle non-subpaths in rule file check

makubacki:
In IgnoreRule.match(), simply return False if the given absolute path is not a subpath of the base path instead of not handling the exception and breaking the overall build.

### DIFF
--- a/edk2toollib/gitignore_parser.py
+++ b/edk2toollib/gitignore_parser.py
@@ -181,7 +181,13 @@ class IgnoreRule(collections.namedtuple("IgnoreRule_", IGNORE_RULE_FIELDS)):
         """Returns True or False if the path matches the rule."""
         matched = False
         if self.base_path:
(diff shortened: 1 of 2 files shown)


edk2toollib/gitignore_parser.py
```diff
@@ -179,7 +179,13 @@ def match(self, abs_path: str) -> bool:
         """Returns True or False if the path matches the rule."""
         matched = False
         if self.base_path:
-            rel_path = _normalize_path(abs_path).relative_to(self.base_path).as_posix()
+            try:
+                rel_path = _normalize_path(abs_path).relative_to(self.base_path).as_posix()
+            except ValueError as e:
```
Javagedes, line 184: Please write a test for this, to ensure the functionality is not accidently lost :)

makubacki: Added

Javagedes (approved):

PR #646, opened 2024-09-30
Fix ruff issues [Rebase & FF]

makubacki:
**pyproject.toml: Update ruff deprecated settings**

Addresses these deprecated messages from ruff:

```
warning: The top-level linter settings are deprecated in favour of
their counterparts in the `lint` section. Please update the
following options in `pyproject.toml`:

  - 'ignore' -> 'lint.ignore'
  - 'select' -> 'lint.select'
  - 'flake8-annotations' -> 'lint.flake8-annotations'
  - 'pydocstyle' -> 'lint.pydocstyle'
```

---

**Fix ruff issues**

Runs `ruff format .` on the workspace.

---

**test.unit: Prevent bare except ruff errors**

Allows ruff to pass in the repo with no problem. Since these are unit
tests, no investment is made to change behavior of the exception
handler.

---

Note: @Javagedes is going to follow up with a PR to update additional documentation and checks for using `ruff format` moving forward.


makubacki:
@Javagedes, it seems strange this wasn't done already. Was there a reason?

Javagedes:
> @Javagedes, it seems strange this wasn't done already. Was there a reason?

No reason - A failure on my part.

Also, it looks like you ran ruff on the entire workspace instead of `ruff edk2toollib`. I'm not against it - buy I should come back with another PR afterwords to update to run ruff on the entire workspace, maybe.

Javagedes:
> @Javagedes, it seems strange this wasn't done already. Was there a reason?

@makubacki you ran `ruff` on the workspace instead of just `edk2toollib` (Which is fine if you want to start doing that instead).

Also, I'm not sure why the single to double quotes rule was applied. That is [Q000](https://docs.astral.sh/ruff/rules/bad-quotes-inline-string/) and I don't have that rule set enabled.

makubacki:
> > @Javagedes, it seems strange this wasn't done already. Was there a reason?
> 
> @makubacki you ran `ruff` on the workspace instead of just `edk2toollib` (Which is fine if you want to start doing that instead).
> 
> Also, I'm not sure why the single to double quotes rule was applied. That is [Q000](https://docs.astral.sh/ruff/rules/bad-quotes-inline-string/) and I don't have that rule set enabled.

That makes sense. I'd prefer we just run it on the whole workspace. It better encompasses the goals of formatting imo to have all the code consistent. I'm also trying to figure the quote thing out.

makubacki:
Pushed two times without realizing the last commit was silently undoing what I kept amending into the second commit to fix CI. It should actually work now.

Javagedes (approved):

PR #652, opened 2024-10-14
Python support: drop 3.10, add 3.13

Javagedes:
## Breaking Change

This commit removes support for python 3.10. Attempting to install v0.22.0 of edk2-pytool-library using a python version less than 3.11 will result in an error from pip along the lines of `ERROR: Could not find a version that satisfies the requirement ...`

## Integration Instructions

If your platform cannot upgrade to 3.11 or above, continue using a version of edk2-python-library less than v0.22.0. Otherwise upgrade your version of platform to 3.11 or greater.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ dynamic = ["version"]
 description = "Python library supporting UEFI EDK2 firmware development"
 readme = {file = "readme.md", content-type = "text/markdown"}
 license = {file = "LICENSE"}
-requires-python = ">=3.10"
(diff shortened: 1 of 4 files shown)


makubacki (approved):

pyproject.toml
makubacki, line 1: Please at least put a breaking change notification in the PR description.

makubacki: I know it's kind of obvious, but still helpful to explicitly state it for someone that might run into a problem in the future.

apop5 (approved):

PR #672, opened 2024-11-21
Update GetEdk2RelativePathFromAbsolutePath and GetAbsolutePathOnThisSystemFromEdk2RelativePath path input type hinting

antklein:
For GetEdk2RelativePathFromAbsolutePath change the type hinting of abspath from Tuple of strings to just str.
For GetAbsolutePathOnThisSystemFromEdk2RelativePath change the type hinting of relpath from Tuple of strings to just str.

This fixes the error messages from Pylint and Ruff about incompatible input types when using the functions as suggested. When attempting to pass a Tuple of strings to the functions, the functions would raise an error.

```
File "...\site-packages\edk2toollib\uefi\edk2\path_utilities.py", line 224, in GetAbsolutePathOnThisSystemFromEdk2RelativePath
    relpath = Path(*[part.replace("\\", "/") for part in relpath])
                     ^^^^^^^^^^^^
AttributeError: 'tuple' object has no attribute 'replace'
```

### DIFF
--- a/edk2toollib/uefi/edk2/path_utilities.py
+++ b/edk2toollib/uefi/edk2/path_utilities.py
@@ -16,7 +16,7 @@ import errno
 import logging
 import os
 from pathlib import Path
-from typing import Iterable, Optional, Tuple
+from typing import Iterable, Optional
 
 
 class Edk2Path(object):
@@ -144,7 +144,7 @@ class Edk2Path(object):
         """List of package paths as strings."""
         return [str(p) for p in self._package_path_list]
 
-    def GetEdk2RelativePathFromAbsolutePath(self, *abspath: Tuple[str, ...]) -> str:
+    def GetEdk2RelativePathFromAbsolutePath(self, *abspath: str) -> str:
         """Transforms an absolute path to an edk2 path relative to the workspace or a packages path.
 
         Args:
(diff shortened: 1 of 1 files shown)


Javagedes:
@antklein What I was trying to annotate with this particular type hint was that you can do things like `GetAbsolutePathOnThisSystemFromEdk2RelativePath("path", "to", "my", "file")`. Is the current type hinting you are changing it to reflect that?

antklein:
> @antklein What I was trying to annotate with this particular type hint was that you can do things like `GetAbsolutePathOnThisSystemFromEdk2RelativePath("path", "to", "my", "file")`. Is the current type hinting you are changing it to reflect that?

Yes, my understanding is that the asterisk handles the fact that the path input can be either be a single or multiple strings, and the type hinting only needs to specify the type of a single input.
https://stackoverflow.com/questions/37031928/type-annotations-for-args-and-kwargs

Javagedes:
> > @antklein What I was trying to annotate with this particular type hint was that you can do things like `GetAbsolutePathOnThisSystemFromEdk2RelativePath("path", "to", "my", "file")`. Is the current type hinting you are changing it to reflect that?
> 
> Yes, my understanding is that the asterisk handles the fact that the path input can be either be a single or multiple strings, and the type hinting only needs to specify the type of a single input. https://stackoverflow.com/questions/37031928/type-annotations-for-args-and-kwargs

Great. Thanks for verifying that for me :)

Javagedes (approved):

PR #684, opened 2025-01-03
base_parser.py: Update precedence of `==` and `!=`

ponchofigueroa:
To align with industry standard order of operations, and with EDKII parsers, equals (`==`) and not equals (`!=`) should have a higher precedence than logical OR (`||`) and logical AND (`&&`) operators.

Prior to this commit, they shared the same precedence level which would lead to evaluations similar to `!if RACECAR == RACECAR || RACECAR == STREETCAR` being completed in the wrong order as compared to EDKII parsers. Additionally, this removes the parentheses that were inserted at the beginning and end of tokens as they appear to serve no further purpose.


Javagedes (changes requested):
@ponchofigueroa to resolve the current CI check, simply run `ruff format .` at the workspace root of the repository (You may have to `pip install ruff`).

As for your changes, I've manually tested your test cases against the EDKII build parsers and the results are the same, so I appreciate you aligning us with them.

For brevity's sake, I would ask that you add a few more tests that use `AND` and `OR` instead of `||` and `&&` and tests using `!=` instead of `==`. 

We have had issues in the past of small changes in our conditional parser breaking scenarios with no tests, so I'd like to make sure we have tests for everything.

I have confidence in your changes not breaking the scenarios I'm asking for additional tests in - This is to ensure future changes do not break these scenarios 😄

ponchofigueroa:
Thanks @Javagedes, I've submitted a new version of the change that essentially recreates a truth table for each for the test I previously added. The new functions are a bit long for my liking, let me know if you think I went overboard and should simplify it somehow.

Do you have any comments on my removal of the parentheses at the start of `_ConvertTokensToPostFix()`? The change is unrelated but looked like a low hanging fruit. While inspecting the code I noticed the input list `tokens` is being modified, so when the function exits the list will have an extra `)` at the end which is generally not a good practice. I believe the only reason to append `)` is due to the stack being initialized with `(`, but I don't see the need to enclose the full expression in parentheses, so I removed them altogether.

Javagedes:
> Thanks @Javagedes, I've submitted a new version of the change that essentially recreates a truth table for each for the test I previously added. The new functions are a bit long for my liking, let me know if you think I went overboard and should simplify it somehow.
> 
> Do you have any comments on my removal of the parentheses at the start of `_ConvertTokensToPostFix()`? The change is unrelated but looked like a low hanging fruit. While inspecting the code I noticed the input list `tokens` is being modified, so when the function exits the list will have an extra `)` at the end which is generally not a good practice. I believe the only reason to append `)` is due to the stack being initialized with `(`, but I don't see the need to enclose the full expression in parentheses, so I removed them altogether.

I also don't see any need for the `(` and `)` anymore. My guess is that it was a remnant of a previous implementation that is no longer needed as improvements were made to the parser. I'm fine with removing it. All of our tests pass. I'll keep an eye out for anyone reporting issues and can always revert that portion of the change.

Thanks for adding the extra tests, also.

Javagedes (approved):

PR #701, opened 2025-03-05
project: Add a minimum supported version

Javagedes:
Adds the concept of a minimum supported version (MSV) to the project.

The project will continue to always maintain and support N-2 versions of Python, however the MSV will be the oldest version of python that can run the project (i.e. the project does not use any features that are newer than the MSV). This value is subject to change at any time due to the needs of the project.


makubacki (approved):

mdkinney (approved):

Javagedes:
@makubacki This PR will also need an equivalent in edk2-pytool-extensions, but I'll make this my priority when I get into office today

PR #710, opened 2025-03-18
locate_tools: QueryVcVariables: Allow exact vc

Javagedes:
Allows a caller to specify the exact compiler version to use for querying variables. This is useful because a system may have a single installation of (example) VS2022, but may have multiple versions of the VS2022 compiler installed. Prior to this change, QueryVcVariables would return variables arbitrarily based off of one of the installed compilers.

This change allows the caller to specify the exact version of the compiler to use for querying variables, so that the returned variables are consistent.

This allows for a bug regarding [WindowsVsToolChain](https://github.com/microsoft/mu_basecore/blob/dev/202405/BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py) to be fixed as specified in: https://github.com/microsoft/mu_basecore/issues/1308

### DIFF
--- a/edk2toollib/windows/locate_tools.py
+++ b/edk2toollib/windows/locate_tools.py
@@ -198,7 +198,7 @@ def QueryVcVariables(
         arch (:obj:`str`, optional): arch to run
         product (:obj:`str`, optional): values defined by vswhere.exe
         vs_version (:obj:`str`, optional): helper to find version of supported VS version (example vs2019)
-        vc_version (:obj:`str`, optional): The exact compiler version to consider for searching for keys
+        vc_version (:obj:`str`, optional): Exact compiler version to consider for variables (example 14.42.34433)
 
     Returns:
         (dict): the appropriate environment variables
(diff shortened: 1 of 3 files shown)


apop5 (approved):

edk2toollib/windows/locate_tools.py
makubacki, line 1: I think this change is useful, but multiple versions can lead to other issues outside of this code. Do you think at least a warning message should be printed? That might help point people in the right direction if something else does come up (particularly when incrementally building across different versions).

makubacki (approved):

edk2toollib/windows/locate_tools.py
Javagedes: I do think a warning would be useful. But I don't think that is appropriate here. I'm not sure I can even tell, at least not anywhere in this exact PR.

I think the best place to do it would be here around here [_get_vc_version](https://github.com/microsoft/mu_basecore/blob/324c815fe78b9fc1359fc26db1bcc14ff46c0762/BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py#L376) if the directory we check has more than one entry.

PR #711, opened 2025-03-19
Optimize `Edk2Path` [Rebase & FF]

makubacki:
**test_path_utilities: Remove debug message count assumption**

Test that a nested package is found by the strings in the debug
message as opposed to the number of messages to avoid the test
being broken when unrelated debug messages are added.

---

**path_utilities: Optimize `Edk2Path` constructor**

`Edk2Path` objects are created relatively frequently throughout a
complete build process. This reduces the average execution time
in a full workspace by approximately half by using `os.walk()`
which is several seconds faster in the workspace and parallelizing
across package paths.

### DIFF
--- a/edk2toollib/uefi/edk2/path_utilities.py
+++ b/edk2toollib/uefi/edk2/path_utilities.py
@@ -12,9 +12,11 @@ the OS specific path with the exception of of any function that returns an Edk2 
 which will always return Posix form.
 """
 
+import concurrent.futures
 import errno
 import logging
 import os
+import timeit
 from pathlib import Path
 from typing import Iterable, Optional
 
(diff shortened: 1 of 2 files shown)


Javagedes:
I did a bunch of work to refactor plugins to reuse the builder created edk2path instead of creating their own. Where are you seeing more instantiations? It would probably be worth a refactor there.

I'm all for this change too, of course.

makubacki:
> I did a bunch of work to refactor plugins to reuse the builder created edk2path instead of creating their own. Where are you seeing more instantiations? It would probably be worth a refactor there.
> 
> I'm all for this change too, of course.

That probably helped, I've only seen about two up to kicking off the build. In mu_tiano_platforms, Edk2Path instantiatiion goes down from ~11s to ~5.5s. On a side note, gathering environment files is ~3s there. I have some perf markers I'd also like to add to edk2-pytool-extensions. But this was the obvious area for improvement at a quick glance.

Javagedes (approved):

PR #717, opened 2025-03-24
Remove the sorted action for candidate_package_path_list

MarcChen46:
Remove the sorted action for candidate_package_path_list to align with original build behavior.

### DIFF
--- a/edk2toollib/uefi/edk2/path_utilities.py
+++ b/edk2toollib/uefi/edk2/path_utilities.py
@@ -97,7 +97,7 @@ class Edk2Path(object):
         end_time = timeit.default_timer()
         self.logger.log(logging.DEBUG, f"Time to check package paths: {end_time - start_time}")
 
-        self._package_path_list = sorted(candidate_package_path_list)
+        self._package_path_list = candidate_package_path_list
 
         if invalid_pp and error_on_invalid_pp:
             raise NotADirectoryError(errno.ENOENT, os.strerror(errno.ENOENT), invalid_pp)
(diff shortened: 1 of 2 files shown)


MarcChen46:
@apop5 @makubacki, please help reviewing this PR

edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -97,7 +97,7 @@ def __init__(
         end_time = timeit.default_timer()
         self.logger.log(logging.DEBUG, f"Time to check package paths: {end_time - start_time}")
 
-        self._package_path_list = sorted(candidate_package_path_list)
+        self._package_path_list = candidate_package_path_list
```
makubacki, line 100: If there is a requirement, it needs to be explicitly stated with a unit test that confirms its behavior.

makubacki: @MarcChen46, we discussed this offline a bit further. Preserving the package list order is important and a documented expectation. Since this line directly impacts the publicly reported package path list, it should not modify the order and your change is good.

However, unit tests are still needed to verify the expected behavior.

MarcChen46: Just added the unit test, please review it again.

tests.unit/test_path_utilities.py
```diff
@@ -183,6 +183,20 @@ def test_invalid_pp(self):
         Edk2Path(str(ws), ["bad_pp_path", "bad_pp_path2", "good_path"], error_on_invalid_pp=False)
 
     @unittest.skipUnless(sys.platform.startswith("win"), "requires Windows")
```
apop5, line 185: I think this @unittest.skip is supposed to be for the test below

apop5 (approved):

tests.unit/test_path_utilities.py
MarcChen46: updated

MarcChen46:
@Javagedes, please review this PR, thanks.

Javagedes (approved):

PR #744, opened 2025-05-20
edk2_db: Fix mis-aligned zip for library dependencies

Javagedes:
This commit fixes a bug that results in a zip of two different sized arrays that must be the same size. This mis-alignment results in recursive library class dependencies being associated with the incorrect parent library or component.

This commit fixes the issue by creating two new lists and only ever appending to both, where as previously a premature continue in a loop had been shown to mis-align the lists.


apop5:
I verified that the previously failing scenario, that valid stuart_report data was generated, but it left out the libraries that did not contain a valid FILE_GUID in their INF.

apop5 (approved):

makubacki (approved):

PR #750, opened 2025-06-23
base_parser: Give input variables precedence to local variables

makubacki:
Right now, if a variable is given as an input variable (e.g. a cmd param), and it has a local variable set to `FALSE`, the input input variable is not considered.

At a higher level, this is causing the DSC parser to not consider an input variable for a build value that is set to `TRUE` on the command line, when it is defaulted to `FALSE` in the `[DEFINES]` section of the DSC.

---

**This PR is in draft** to start a conversation about how to handle this. Clearly, the `test_replace_macro_local_var_priority()` unit test expected the previous behavior, but that doesn't align with how the logic in [edk2_pr_eval.py](https://github.com/tianocore/edk2-pytool-extensions/blob/master/edk2toolext/invocables/edk2_pr_eval.py#L322) is using the return value from [`PlatformDscInfo = self.PlatformSettings.GetPlatformDscAndConfig()`](https://github.com/tianocore/edk2-pytool-extensions/blob/master/edk2toolext/invocables/edk2_pr_eval.py#L309).


Javagedes:
I believe the main use case for the current logic is parsing a DSC, taking the variables parsed from the DSC, and passing them to the FDF to parse the FDF. In this scenario, (I would need to read the build tools spec when I get back), it is expected that the local variable defined in the FDF takes precedence over the one defined in the DSC. This change would invert that (IIRC).

makubacki:
For reference, the original commit up to this point was https://github.com/tianocore/edk2-pytool-library/pull/750/commits/52ceec3d8a060f4ff91ea8848b5897d3ebd2b4d6.

I updated to an approach where the current behavior is preserved, but the caller (EDK2 PR Eval Policy 4) can prioritize passed in variables (commit 899a552ec55a40d3eddad3ac00b058f99a3f1bca).

makubacki:
> > For reference, the original commit up to this point was [52ceec3](https://github.com/tianocore/edk2-pytool-library/commit/52ceec3d8a060f4ff91ea8848b5897d3ebd2b4d6).
> > I updated to an approach where the current behavior is preserved, but the caller (EDK2 PR Eval Policy 4) can prioritize passed in variables (commit [899a552](https://github.com/tianocore/edk2-pytool-library/commit/899a552ec55a40d3eddad3ac00b058f99a3f1bca)).
> 
> The way to change precedence looks good to me. Nice and Simple. Is this customization special for PR eval, or is it just a different setting per configuration file (INF, DSC, DEC, FDF). I'm wonder if we should/could just manually set this variable for each subclass instead of requiring the user to override it?

If we accepted this approach, my plan was to set it such that the user would not need to do anything.

makubacki:
Switched back to the original commit per offline discussion.

Javagedes (approved):

tests.unit/parsers/test_base_parser.py
```diff
@@ -39,7 +39,7 @@ def test_replace_macro_local_var_priority(self):
         parser.SetInputVars({"name": "sean"})
         parser.LocalVars["name"] = "fred"
         line = "Hello $(name)!"
-        self.assertEqual(parser.ReplaceVariables(line), "Hello fred!")
+        self.assertEqual(parser.ReplaceVariables(line), "Hello sean!")
```
spbrogan, line 42: nice touch

spbrogan (approved):
looks good.

makubacki:
@Javagedes, I think this can be merged when ready.

PR #756, opened 2025-08-20
pyproject.toml: Update setuptools to latest (80.9.0)

makubacki:
Update the minimum setuptools version to the latest version (80.9.0) to ensure PYSEC-2025-49 is mitigated.

https://osv.dev/vulnerability/PYSEC-2025-49


Javagedes:
I've validated we can continue to generate the wheel and sdist using the updated setuptools. Thanks!

PR #758, opened 2025-09-21
GetContainingPackage() returns relative path

yangrongwei:
Return the **relative path** rather than the **leaf name** of the package path.

The **leaf name** results in unnecessary paths being added to PACKAGES_PATH, solely to satisfy the subsequent call to `GetAbsolutePathOnThisSystemFromEdk2RelativePath()`.

### DIFF
--- a/edk2toollib/uefi/edk2/path_utilities.py
+++ b/edk2toollib/uefi/edk2/path_utilities.py
@@ -264,7 +264,7 @@ class Edk2Path(object):
 
         return None
 
-    def GetContainingPackage(self, InputPath: str) -> str:
+    def GetContainingPackage(self, InputPath: str) -> str | None:
         """Finds the package that contains the given path.
 
         This isn't perfect, but at least identifies the direcotry consistently.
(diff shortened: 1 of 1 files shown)


Javagedes:
I would also be open to moving this code to a new function called `GetContainingPackageRelativePath`

Then the existing `GetContainingPackage` simply calls GetContainingPackage and returns the folder name.

yangrongwei:
> [Javagedes](/Javagedes)

Changing the return is safer than you feared.
Adding a new function is messier than you think.

Take the edk2 repo as an example, almost all plug-in call `GetAbsolutePathOnThisSystemFromEdk2RelativePath()` after `GetContainingPackage()`.  This is a huge effort to change every plug-in.

Current platform owners will not be affected by this change. They provided a superset of PACKAGES_PATH compared to what was expected.

The risk of the order in PACKAGES_PATH is even lower now, because this function returns a more precise relative path.

New platforms will benefit from this change. Since they don’t need to customize PACKAGES_PATH specifically for Stuart, migrating from the `build.py` based build to Stuart will be easier.

The new `GetContainingPackageRelativePath` will not provide this benefit unless all plug-ins are migrated to use this new function.

Yours,
Raymond Yang

PR #789, opened 2026-03-13
Fix build report parsing for long files

NishanthSanjeevi:
## Description
Fix build report parsing for long file names

## How This Was Tested
Errors no longer observed while building files with long names


Javagedes (changes requested):
Please provide additional information in the pull request explaining the exact issue this is facing with example, and also please write at least one unit test that was broken before this change and is fixed with this change to prevent regressions in the future.

Thanks!

Javagedes:
@NishanthSanjeevi Do you plan on continuing to work on this issue or can it be closed?

PR #798, opened 2026-04-07
edk2toollib: Fix Pcds are not parsed when next to symbol

PaddyDengAmi:
The original string.split in ReplacePcds yelds incorrect result when the pcd is right next to symbols.

Sign-off-by: Paddy Deng <paddydeng@ami.com>


Javagedes (approved):

Javagedes:
This looks great. Thank you for the contribution!

PR #828, opened 2026-08-31
dsc_parser: Honor DEFINE statements outside [Defines]

r3wretrhy:
## Description

`DscParser` only recorded `DEFINE` assignments while `CurrentSection` was `DEFINES` or `BUILDOPTIONS`. The EDK II DSC spec (2.2.6) allows `DEFINE` in other sections and scopes them to the section they appear in; only `[Defines]` macros are global. A `DEFINE` under `[LibraryClasses]` (or `[BuildOptions]`) was either dropped or stuffed into `LocalVars`, so later `!ifdef` / `$(MACRO)` use was wrong.

This records DEFINE-prefixed assignments in any section:
- `[Defines]` → `LocalVars` (global)
- `[BuildOptions]` and **all other sections** → `SectionMacros` keyed by `(section, arch, module)`, with common-arch / empty-module matching more specific subsections of the same section type

DEFINE lines are skipped in `_parse_libraries` instead of being logged as unexpected library rows.

Fixes #788

## Testing

- `python -m pytest tests.unit/parsers/test_dsc_parser.py` (15 passed)
- `python -m pytest tests.unit/parsers/` (154 passed)
- `python -m pytest tests.unit` (403 passed, 21 skipped)

New coverage:
- BuildOptions DEFINE is not global; common→X64 visible; X64-local does not leak to IA32
- common vs arch override precedence (`ROOT` CommonPkg vs X64Pkg → `X64.PEIM` picks X64Pkg)

### DIFF
--- a/edk2toollib/uefi/edk2/parsers/dsc_parser.py
+++ b/edk2toollib/uefi/edk2/parsers/dsc_parser.py
@@ -60,6 +60,10 @@ class DscParser(HashFileParser):
         self.PcdValueDict = {}
         self._no_fail_mode = False
         self._dsc_file_paths = set()  # This includes the full paths for every DSC that makes up the file
+        # DEFINE outside [Defines] is scoped to that section (DSC spec 2.2.6).
+        # Key: (section_type, arch, module)  e.g. ("LIBRARYCLASSES", "X64", "PEIM")
+        self.SectionMacros: dict[tuple[str, str, str], dict[str, str]] = {}
+        self.CurrentSectionScopes: list[tuple[str, str, str]] = []
 
     def ReplacePcds(self, line: str) -> str:
         """Attempts to replace a token if it is a PCD token."""
@@ -68,6 +72,68 @@ class DscParser(HashFileParser):
                 if token in self.PcdValueDict:
                     line = line.replace(token, self.PcdValueDict[token])
         return line
+
+    def _parse_section_scopes(self, header_line: str) -> list[tuple[str, str, str]]:
+        """Split a [Section.Arch.Module, ...] header into scope tuples."""
+        inner = header_line.strip()
+        if inner.startswith("["):
+            inner = inner[1:]
+        if inner.endswith("]"):
+            inner = inner[:-1]
+        scopes: list[tuple[str, str, str]] = []
+        for part in inner.split(","):
+            part = part.strip()
+            if not part:
+                continue
+            bits = [b.strip() for b in part.split(".")]
+            stype = bits[0].upper()
+            arch = bits[1] if len(bits) > 1 else "common"
+            module = bits[2].upper() if len(bits) > 2 else ""
+            scopes.append((stype, arch, module))
+        return scopes
+
+    def _scope_applies(
+        self, defined: tuple[str, str, str], current: tuple[str, str, str]
+    ) -> bool:
+        """True if a DEFINE stored under *defined* is visible in *current*.
+
+        common arch matches every arch. An empty module matches every module
+        type. A more specific DEFINE does not leak into a sibling or parent.
+        """
+        dstype, darch, dmod = defined
+        cstype, carch, cmod = current
+        if dstype != cstype:
+            return False
+        if darch.lower() != "common" and darch.lower() != carch.lower():
+            return False
+        if dmod and dmod != cmod:
+            return False
+        return True
+
+    def _scope_specificity(self, scope: tuple[str, str, str]) -> tuple[int, int]:
+        stype, arch, module = scope
+        return (0 if arch.lower() == "common" else 1, 1 if module else 0)
+
+    def _section_macro_value(self, token: str) -> Optional[str]:
+        found: Optional[str] = None
+        best = (-1, -1)
+        for current in self.CurrentSectionScopes:
+            for defined, macros in self.SectionMacros.items():
+                if token not in macros or not self._scope_applies(defined, current):
+                    continue
+                spec = self._scope_specificity(defined)
+                if spec >= best:
+                    best = spec
+                    found = macros[token]
+        return found
+
+    def _FindReplacementForToken(self, token: str, replace_if_not_found: bool = False) -> str:
+        v = super()._FindReplacementForToken(token, False)
+        if v is None:
+            v = self._section_macro_value(token)
+        if v is None and replace_if_not_found:
+            v = self._MacroNotDefinedValue
+        return v
 
     def __ParseLine(self, Line: str, file_name: Optional[str] = None, lineno: int = None) -> tuple:
         line_stripped = self.StripComment(Line).strip()
@@ -105,6 +171,7 @@ class DscParser(HashFileParser):
         (IsNew, Section) = self.ParseNewSection(line_resolved)
         if IsNew:
             self.CurrentSection = Section.upper()
+            self.CurrentSectionScopes = self._parse_section_scopes(line_resolved)
             self.Logger.debug("New Section: %s" % self.CurrentSection)
             self.Logger.debug("FullSection: %s" % self.CurrentFullSection)
             return (line_resolved, [], None)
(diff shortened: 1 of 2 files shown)


Javagedes:
@r3wretrhy I do agree that the current implementation is wrong (as noted in #788 that this PR is linked to).

This pull request updates the logic to add any macro definition to our global variable store (`self.LocalVars`). This itself is a bug because per the DSC specification (2.2.6):

> Macro Definition statements that appear within a section of the file (other than the [Defines] section) are scoped to the section they are defined in. If the Macro statement is within the [Defines] section, then the Macro is considered global to the entire DSC file

To properly resolve #788, a solution would need to account for the fact that any variables set outside the `Defines` section is only applicable in an equivalent section, while also considering if it should be applied based off the subsection - `[<section>.<arch>.<module>]`.

The below example comes from the DSC specification and shows some of the complexity that needs to be considered to make a full fix for this issue.

```
[LibraryClasses.common]
  DEFINE MDE = MdePkg/Library
  BaseLib|$(MDE)/BaseLib.inf
[LibraryClasses.X64, LibraryClasses.IA32]
  # Can use $(MDE), cannot	use	$(MDEMEM)
  DEFINE PERF = PerformancePkg/Library
  TimerLib|$(PERF)/DxeTscTimerLib/DxeTscTimerLib.inf
[LibraryClasses.X64.PEIM]
  #  Can use $(MDE) and $(PERF)
  DEFINE MDEMEM = $(MDE)/PeiMemoryAllocationLib
  MemoryAllocationLib|$(MDEMEM)/PeiMemoryAllocationLib.inf
[LibraryClasses.IPF]
  # Cannot use $(PERF) or $(MDEMEM)
  # Can use $(MDE) from the common section
  PalLib|$(MDE)/UefiPalLib/UefiPalLib.inf
  TimerLib|$(MDE)/BaseTimerLibNullTemplate/BaseTimerLibNullTemplate.inf
```

r3wretrhy:
You're right: stuffing those DEFINE values into LocalVars made them global.

They are now stored per (section, arch, module). `common` arch matches every arch, and a DEFINE without a module type matches more specific subsections of the same section. A DEFINE in `[LibraryClasses.common.PEI_CORE]` does not apply to `DXE_DRIVER`. `[Defines]` / `[BuildOptions]` stay on LocalVars.

The tests now follow the DSC 2.2.6 example (MDE / PERF / MDEMEM visibility), including the comma-separated `[LibraryClasses.X64, LibraryClasses.IA32]` header.

r3wretrhy:
Thanks @Javagedes — agreed that stuffing section DEFINEs into `LocalVars` was wrong.

Per DSC 2.2.6, macros outside `[Defines]` are scoped to the section they appear in (common arch matches every arch; an empty module modifier matches more specific subsections of the same section type). `[Defines]` / `[BuildOptions]` remain global on `LocalVars`.

This update stores non-`[Defines]` DEFINE values in `SectionMacros` keyed by `(section, arch, module)` and resolves them through `_FindReplacementForToken`, matching the MDE / PERF / MDEMEM example from the spec. Rebased onto current `master` (`8ed892c`).

r3wretrhy:
CI on `70df421` failed `ruff format --check` and `cspell` on the DEFINE-scoping helpers added in this PR.

Amended in `0e45add`:
- `ruff format` on `dsc_parser.py`
- renamed short locals in `_scope_applies` (`dstype`/`darch`/…) so cspell is clean
- added `MDEMEM` to `.cspell.json` (macro name from the DSC 2.2.6 example used in the new tests)

Local checks: `ruff format --check` / `ruff check` clean on the touched files; `cspell` clean on those two files; `pytest tests.unit/parsers/` 154 passed.

Javagedes:
Hi @r3wretrhy, I will get this reviewed today to provide feedback or to get it merged in. Thank you for this work!

edk2toollib/uefi/edk2/parsers/dsc_parser.py
```diff
@@ -70,6 +74,66 @@ def ReplacePcds(self, line: str) -> str:
...
+                if token not in macros or not self._scope_applies(defined, current):
+                    continue
+                spec = self._scope_specificity(defined)
+                if spec >= best:
+                    best = spec
+                    found = macros[token]
+        return found
+
+    def _FindReplacementForToken(self, token: str, replace_if_not_found: bool = False) -> str:
+        v = super()._FindReplacementForToken(token, False)
```
Javagedes: This appears to be backwards. A macro defined in a section should take presidence over one defined in the `[Defines]` section, but this appears to be the opposite.

e.g. 

```
[Defines]
  DEFINE ROOT = GlobalPkg

[LibraryClasses]
  DEFINE ROOT = LocalPkg
  BaseLib|$(ROOT)/BaseLib.inf
```

I would expected ROOT to be LocalPkg. Please also add a test for this scenario.

edk2toollib/uefi/edk2/parsers/dsc_parser.py
```diff
@@ -231,31 +296,41 @@ def __ParseDefineLine(self, Line: str) -> tuple:
...
+            for scope in self.CurrentSectionScopes:
+                self.SectionMacros.setdefault(scope, {})[left] = right
+            self.Logger.debug("Section DEFINE %s = %s scopes=%s" % (left, right, self.CurrentSectionScopes))
             return (line_resolved, [])
 
+        self.LocalVars[left] = right
+        self.Logger.debug("Key,values found:  %s = %s" % (left, right))
+
+        for var in self.LocalVars:
+            self.LocalVars[var] = self.ReplaceVariables(self.LocalVars[var])
```
Javagedes, line 372: The parser currently performs a complete DEFINE pre-pass (__ProcessDefines() from ParseFile()) before parsing and resolving ordinary lines with __ProcessMore(). This reduces the definitions for each section scope to one final macro value before the full parse begins.

That approach is incompatible with section-local DEFINE semantics. According to the DSC macro rules, a section-local definition applies to entries following that definition, and the most recent definition takes precedence from that point forward.

During the pre-pass, multiple definitions of the same macro in the same scope overwrite the same SectionMacros entry. Because that map is retained for the full parsing pass, the final definition is visible from the beginning of the section and is applied retroactively.

For example:

```
[LibraryClasses]
  DEFINE ROOT = InitialPkg
  BeforeLib|$(ROOT)/BeforeLib.inf
  DEFINE ROOT = LocalPkg
  NextLib|$(ROOT)/NextLib.inf
  DEFINE ROOT = ExternalPkg
  FinalLib|$(ROOT)/FinalLib.inf
```

The expected resolved library instances are:

```
[LibraryClasses]
BeforeLib|InitialPkg/BeforeLib.inf
NextLib|LocalPkg/NextLib.inf
FinalLib|ExternalPkg/FinalLib.inf
```

With this implementation, they are:

```
BeforeLib|ExternalPkg/BeforeLib.inf
NextLib|ExternalPkg/NextLib.inf
FinalLib|ExternalPkg/FinalLib.inf
```

Section-local definitions therefore need to be activated in source order during the full parsing pass—or otherwise stored with their source position—rather than reduced to one final value before parsing begins.

edk2toollib/uefi/edk2/parsers/dsc_parser.py
```diff
@@ -70,6 +74,66 @@ def ReplacePcds(self, line: str) -> str:
...
+            return False
+        if defined_mod and defined_mod != current_mod:
+            return False
+        return True
+
+    def _scope_specificity(self, scope: tuple[str, str, str]) -> tuple[int, int]:
+        stype, arch, module = scope
+        return (0 if arch.lower() == "common" else 1, 1 if module else 0)
+
+    def _section_macro_value(self, token: str) -> Optional[str]:
```
Javagedes, line 116: Unfortuantely selecting the right macro for the right section has become a whole lot more complex with these changes. Particularly when it comes to evaluating sections with multiple inner grouping (like `[Components.IA32, Components.X64]`).

This current implementation of `_section_macro_value` selects the last most specific option for the macro value, even when there are multiple scopes. So in the above example, even if both `Components.IA32` and `Components.X64` have a valid scope, the `Components.X64` scope will always set the variable.

Unforuntely this will require a fairly drastic change to parsing to be proper. Consider the following:

```
[Components.IA32]
DEFINE MY_VAR = "SomePkg"

[Components.X64]
DEFINE MY_VAR = "OtherPkg"

[Components.IA32, Components.X64]
  $(MY_VAR)/MyDriver/MyDriver.inf

With your current example, the below will always be evaluated as `OtherPkg/MyDriver/MyDriver.inf` and will always be registered to the `self.ThreeMods`. This is a bug due to this particular function, but also due to the `self.ParseNewSection`'s implementation in `base_parser.py`.

However what should be happening is we should be evaluating it twice. Once as `SomePkg/MyDriver/MyDriver.inf` that is registered to self.ThreeMods and once as `OtherPkg/MyDriver/MyDriver.inf` that is registered to self.SixMods.

edk2toollib/uefi/edk2/parsers/dsc_parser.py
r3wretrhy: Fixed in `44d8023`: `_FindReplacementForToken` now prefers `SectionMacros` over `[Defines]` / `LocalVars`, and added a regression test for the `ROOT = GlobalPkg` vs section `ROOT = LocalPkg` case.

edk2toollib/uefi/edk2/parsers/dsc_parser.py
r3wretrhy: Agreed — the DEFINE pre-pass was collapsing each scope to a final value before INF lines were resolved.

In `44d8023` the pre-pass only materializes `[Defines]` into `LocalVars`. Section-scoped DEFINEs (including `[BuildOptions]`) are applied in source order during the full parse, and `ResetParserState` clears `SectionMacros` between passes. Added a regression test for the InitialPkg → LocalPkg → ExternalPkg sequence.

edk2toollib/uefi/edk2/parsers/dsc_parser.py
r3wretrhy: Addressed in `44d8023` without changing `base_parser.ParseNewSection`:

- Component INF lines under a multi-scope header are resolved once per `CurrentSectionScopes` entry and routed to `ThreeMods` / `SixMods` / `OtherMods` for that arch.
- `Lines` keep unresolved section macros so `_parse_components` can expand `$(MY_VAR)` per arch (and DEFINE lines in Components are skipped).

Added a regression test for the IA32=`SomePkg` / X64=`OtherPkg` example you gave. Happy to iterate if you still want a deeper `ParseNewSection` rewrite.

r3wretrhy:
Thanks @Javagedes — addressed the three review notes in `44d8023`:

1. **Section DEFINE overrides `[Defines]`** — `_FindReplacementForToken` prefers `SectionMacros` over `LocalVars`.
2. **Source-order DEFINE** — section DEFINEs are no longer finalized in the pre-pass; they activate during the full parse so earlier INF lines keep the DEFINE that was in effect at that point.
3. **Multi-arch Components headers** — INF lines under `[Components.IA32, Components.X64]` are evaluated once per scope (IA32 → `ThreeMods` / X64 → `SixMods`), and `_parse_components` expands remaining section macros per arch.

Local checks: `ruff format` / `ruff check` clean on the touched files; `cspell` clean; `pytest tests.unit/parsers/` 157 passed (including three new regressions for the cases above).

r3wretrhy:
CI `BasicDevTests.py` failed on `44d8023` with an ASCII encoding error in `dsc_parser.py` (Unicode `→` in a docstring at byte 6762).

Fixed in `af93f23`: replaced those arrows with ASCII `->` in the `dsc_parser.py` docstring and the two matching test docstrings. No logic change.

Local check: `pytest tests.unit/parsers/` → 157 passed; both touched `.py` files decode as pure ASCII.

PR #829, opened 2026-08-31
dsc_parser: Restore TargetFilePath after nested !include

### LINKED ISSUE #394: [Bug]: `base_parser.TargetFilePath` not reverted when exiting an !include parse

r3wretrhy:
## Description

`FindPath` resolves `!include` first relative to the file currently being parsed. After reading an include, `TargetFilePath` stayed on that include, so a second sibling include was joined onto the include directory and failed.

This pushes and pops `TargetFilePath` around include expansion so sibling relative includes resolve from the parent again.

Fixes #394

## Testing

- `python -m pytest tests.unit/parsers/test_dsc_parser.py::TestDscParserIncludes`
- `python -m pytest tests.unit` (414 passed, 4 skipped)

### DIFF
--- a/edk2toollib/uefi/edk2/parsers/dsc_parser.py
+++ b/edk2toollib/uefi/edk2/parsers/dsc_parser.py
@@ -60,6 +60,7 @@ class DscParser(HashFileParser):
         self.PcdValueDict = {}
         self._no_fail_mode = False
         self._dsc_file_paths = set()  # This includes the full paths for every DSC that makes up the file
+        self._target_file_stack = []  # previous TargetFilePath values for nested !include
 
     def ReplacePcds(self, line: str) -> str:
         """Attempts to replace a token if it is a PCD token."""
@@ -289,6 +290,8 @@ class DscParser(HashFileParser):
                 if len(line) > 0:
                     self.Lines.append(line)
                 self.__ProcessMore(add, file_name=new_file)
+                if add:
+                    self._PopTargetFile()
             except Exception as e:
                 # check if we're in no fail mode or not
                 if not self._no_fail_mode:  # if we are, fail
@@ -314,6 +317,8 @@ class DscParser(HashFileParser):
             try:
                 (line, add) = self.__ParseDefineLine(raw_line)
                 self.__ProcessDefines(add)
+                if add:
+                    self._PopTargetFile()
             except Exception:
                 # Since we're going to do this in ProcessMore, don't warn people if there's an exception
                 # otherwise, raise the exception and act normally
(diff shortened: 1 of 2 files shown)


Javagedes (approved):

Javagedes:
This looks good to me. Thank you for the contribution. I'll hold off on doing a release to see if you want to put more work into #828. If not, I can do a release.

PR #830, opened 2026-09-02
Update EDK2Path class to handle os.PathLike objects correctly for Python 3.14+

antklein:
The change makes `Edk2Path` support `os.PathLike` inputs, as its type annotations promise.

- Workspace paths are slash-normalized only when they are strings.
- Package paths receive the same conditional handling through an explicit loop.

Previously, passing a `pathlib.Path` called its filesystem `replace()` method as though it were `str.replace()`, causing a `TypeError`. Other `PathLike` implementations could raise `AttributeError`.

The fix preserves existing string behavior, including Windows separator support, while allowing `Path` objects to pass safely into `Path(...)`. This aligns runtime behavior with the public API and avoids forcing callers to use `str(path)`. Added a unit test to verify pathlike support.

### DIFF
--- a/edk2toollib/uefi/edk2/path_utilities.py
+++ b/edk2toollib/uefi/edk2/path_utilities.py
@@ -58,11 +58,11 @@ class Edk2Path(object):
         # Other code is dependent the following types, so keep it that way:
         #   - self.PackagePathList: List[str]
         #   - self.WorkspacePath: str
-        ws = ws.replace("\\", "/")
-        workspace_candidate_path = Path(ws)
+        normalized_workspace = ws.replace("\\", "/") if isinstance(ws, str) else ws
+        workspace_candidate_path = Path(normalized_workspace)
 
         if not workspace_candidate_path.is_absolute():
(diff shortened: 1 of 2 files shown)


edk2toollib/uefi/edk2/path_utilities.py
```diff
@@ -58,11 +58,11 @@ def __init__(
         # Other code is dependent the following types, so keep it that way:
         #   - self.PackagePathList: List[str]
         #   - self.WorkspacePath: str
-        ws = ws.replace("\\", "/")
-        workspace_candidate_path = Path(ws)
+        ws_str = ws.replace("\\", "/") if isinstance(ws, str) else ws
```
Javagedes: `ws_str` is whatever type `ws` is, so naming this `ws_str` is a little bit of a misnomer.

Javagedes (approved):

edk2toollib/uefi/edk2/path_utilities.py
antklein: Updated the name to `normalized_workspace`

